PR #2, opened 2021-11-15
Add decompilation test cases

AlviseDeFaveri:


tests/analysis/AnalysisTests.cmake
```diff
@@ -51,3 +51,9 @@ register_artifact("tests_analysis_StackAnalysis" "stack-argument-contradiction"
 
 register_artifact_category("tests_analysis_RestructureCFGPass" "OFF")
 register_artifact("tests_analysis_RestructureCFGPass" "throttle" "x86_64" "throttle.c")
+
+register_artifact_category("tests_analysis_Decompilation" "OFF")
+register_artifact("tests_analysis_Decompilation" "struct" "dynamic_native" "struct.c")
+register_artifact("tests_analysis_Decompilation" "array" "dynamic_native" "array.c")
+register_artifact("tests_analysis_Decompilation" "linked_list" "dynamic_native" "linked_list.c")
+register_artifact("tests_analysis_Decompilation" "array_of_structs" "dynamic_native" "array_of_structs.c")
```
aleclearmind: Missing line at the end of file

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,27 @@
+#include <stdint.h>
+
+struct __attribute__((packed)) X {
+  uint16_t a[10];
+  uint64_t b;
+};
+
+uint16_t __attribute__((noinline)) getA(struct X *x) {
```
aleclearmind: Usually attributes go at the end of a *declaration*. Doesn't this give a warning?
Also, you're using `noinline` basically everywhere. I'd go for `-fno-inline` and be happy.

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,27 @@
+#include <stdint.h>
+
+struct __attribute__((packed)) X {
+  uint16_t a[10];
+  uint64_t b;
+};
```
aleclearmind: Can you come up with some more meaningful naming? Also let's be coherent in coding style.
In C, usually we go like python. Snake case, first letter lowercase.

Something like this?
```c
typedef struct {
  uint16_t uint16_array[10];
  uint64_t uint64;
} X;
```

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,27 @@
...
+  for (int i = 0; i < 10; i++) {
+    uint16_t cur = x->a[i];
+    if (cur > max)
+      max = cur;
+  }
+
+  return max;
+}
+
+uint64_t __attribute__((noinline)) getB(struct X *x) { return (x->b) * 2; }
```
aleclearmind: Let's not have one line functions.
Also drop superfluous parenthesis.

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,27 @@
...
+
+int __attribute__((noinline)) main(int argc, char **argv) {
+  struct X *TheX = *(struct X **)(argv[0]);
```
aleclearmind: Not super nice, can you maybe have a separate function that uses a global variable.
Would that make a difference?

tests/analysis/Decompilation/dynamic_native/linked_list.c
```diff
@@ -0,0 +1,44 @@
...
+}
+
+int __attribute__((noinline)) main(int argc, char **argv) {
+
+  struct Node *First = (struct Node *)malloc(sizeof(struct Node));
+  First->prev = NULL;
+
+  struct Node *Cur = First;
+  for (int i = 0; i < 10; i++) {
+    struct Node *New = (struct Node *)malloc(sizeof(struct Node));
```
aleclearmind: I know this code will never run, but still, any reason to use `malloc` instead of locals?
If you like `malloc`, free them too!

aleclearmind (changes requested):

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: How do I add this flag to the qa pipeline?

tests/analysis/Decompilation/dynamic_native/linked_list.c
AlviseDeFaveri: I want to produce pointers to stuff that is not on the stack, so that the results are not polluted by the quality of previous analyses and the stack might be handled in a special way during decompilation. My goal here is to check the results of type system recovery in the most isolated way possible.
If there are better ways I'm open to suggestions.

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: I can check how the code is generated and decompiled using global variables. I think that the simpler the decompiled code comes out from this, the better, since it is meant to be inspected by hand.

tests/analysis/Decompilation/dynamic_native/array.c
aleclearmind: Something like `set(CFLAGS_CATEGORY_tests_analysis_Decompilation -fno-inline)` in `derived-artifacts/compiled.cmake`.

tests/analysis/Decompilation/dynamic_native/linked_list.c
aleclearmind: Makes sense. Just add the `free`s then.

tests/analysis/Decompilation/dynamic_native/array.c
aleclearmind: I'm just suggesting to have a global variable of type `struct X *` and used that instead of using `argv[0]`.
You will see a load from a fixed address as opposed to using the argument (in the register).

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: 1. Fyi this does not generate a warning, neither with gcc nor with clang. Output example:

```c
 →  gcc -Wall -Wextra -Wpedantic array.c        
array.c: In function ‘main’:
array.c:21:40: warning: unused parameter ‘argc’ [-Wunused-parameter]
   21 | int __attribute__((noinline)) main(int argc, char **argv) {
      |                                    ~~~~^~~~
```

2. Added `-fno-inline` as suggested

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: Only for variable and fields or also for `typedef`s?

tests/analysis/AnalysisTests.cmake
AlviseDeFaveri: fixed

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: fixed

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: I tried to improve naming, please check if it's ok.

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: Added global variables and a getter function that returns the address of the global variable.

tests/analysis/Decompilation/dynamic_native/linked_list.c
AlviseDeFaveri: fixed

derived-artifacts/compiled.cmake
```diff
@@ -55,3 +55,6 @@ macro(artifact_handler CATEGORY INPUT_FILE CONFIGURATION OUTPUT TARGET_NAME)
 
 endmacro()
 register_derived_artifact("compiled" "compiled-run" "" "DIRECTORY")
+
+# Flags for Decompilation tests
+set(CFLAGS_CATEGORY_tests_analysis_Decompilation -fno-inline)
```
aleclearmind: Put these close to the others (see `set(CFLAGS_CATEGORY_tests_runtime...`).

tests/analysis/Decompilation/dynamic_native/linked_list.c
```diff
@@ -0,0 +1,51 @@
+#include <malloc.h>
```
aleclearmind, line 5: Add to all files the copyright notice. See `tests/runtime/floating-point.c`.

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,35 @@
+#include <stdint.h>
+
+typedef struct __attribute__((packed)) {
+  uint16_t uint16_array[10];
+  uint64_t uint64;
+} StructT;
```
aleclearmind: What's the trailing `T` for? Maybe `TestStruct`?

tests/analysis/Decompilation/dynamic_native/array.c
```diff
@@ -0,0 +1,35 @@
...
+
+StructT glob_struct;
```
aleclearmind: Do not abbreviate names ([B.6](https://pad.rev.ng/YwADn1W7QZGKGbY7yaVg_Q#)).

aleclearmind (changes requested):

derived-artifacts/compiled.cmake
AlviseDeFaveri: fixed

tests/analysis/Decompilation/dynamic_native/linked_list.c
AlviseDeFaveri: fixed

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: fixed

tests/analysis/Decompilation/dynamic_native/array.c
AlviseDeFaveri: fixed

aleclearmind:
Merged in 219251a. Thanks!

PR #3, opened 2021-11-17
Add `STT_FUNC` type to functions

antoniofrighetto:
Associate symbols with function type.


aleclearmind:
Looks great! Can you just push `.type` right above the function declaration so next time copy paste stuff around we won't forget about this?

aleclearmind:
Merged, thanks!

PR #4, opened 2021-11-22
Implement the ABI testing artifact generation

Cvelth:


tests/abi/abi.c
```diff
@@ -0,0 +1,100 @@
...
+#define GENERATED_COUNT REGISTER_COUNT + STACK_VALUE_COUNT
+
+void dumpGenerated(REGISTER_C_TYPE *Generated) {
+  /* Dump registers*/
+  puts("Registers:");
+  {
+    uint32_t RegisterIndex = REGISTER_COUNT - 1;
+    #define DUMP_REGISTER(REGISTER) printf("  \"%s\": 0x%016lx\n", \
+                                           STRINGIFY(REGISTER), \
+                                           Generated[RegisterIndex--]);
```
aleclearmind: I'd use `do { } while (0)` syntax for multiline macros everywhere.

tests/abi/abi.c
```diff
@@ -0,0 +1,100 @@
...
+    FOR_EACH(DUMP_REGISTER, REGISTER_LIST);
+    #undef DUMP_REGISTER
+  }
+
+  /* Dump Stack*/
```
aleclearmind: Everywhere
```suggestion
  /* Dump Stack */
```

tests/abi/abi.c
```diff
@@ -0,0 +1,100 @@
+#include <stdint.h>
```
aleclearmind: Copyright notice?

If we don't have it in revng-qa, please add it everywhere (see what revng does for C files).

tests/abi/functions.h
```diff
@@ -0,0 +1,84 @@
+#ifndef __REVNG_QA_ABI_FUNCTIONS_H
+#define __REVNG_QA_ABI_FUNCTIONS_H
```
aleclearmind: I think `__` is a reserved prefix in C. Use `#pragma once`.

tests/abi/abi.c
```diff
@@ -0,0 +1,100 @@
...
+# include "platform_specific/x86_64.h"
+#elif __x86__
+# include "platform_specific/x86.h"
+#else
+ _Static_assert(0, "unsupported platform architecture");
+#endif
+
+/* LFSR generator */
+
+static uint64_t LFSR = 0xdc984ef2647af904ull;
```
aleclearmind: Is this a backdoor? Document how you obtained this value!!!

aleclearmind (changes requested):

tests/abi/abi.c
Cvelth: Wait, the starting point matters? O_o I was acting under the impression that this is pretty much just a seed, so a I picked an arbitrary value.

The only requirement that I know of is that this value should be neither all zeros nor all ones.

tests/abi/functions.h
Cvelth: I REALLY wanted to use `#pragma once` but then I got hesitant about our compiler support. But, ok, since it's no older than `gcc 7.2`, this should be fine, I should have switched right away.

tests/abi/abi.c
Cvelth: [Done](https://github.com/revng/revng-qa/pull/4/commits/cfb308ce4890fb6cc01c353d65ad41be3c0fa26a).

tests/abi/functions.h
Cvelth: [Done](https://github.com/revng/revng-qa/pull/4/commits/cfb308ce4890fb6cc01c353d65ad41be3c0fa26a).

tests/abi/abi.c
Cvelth: [Done](https://github.com/revng/revng-qa/pull/4/commits/62e925ba61a1588cbe4414159485d3f6b425ae40).

tests/abi/abi.c
aleclearmind: I'm joking, but in crypto this is actually an issue:
* https://en.wikipedia.org/wiki/Nothing-up-my-sleeve_number
* https://safecurves.cr.yp.to/rigid.html

Cvelth: Ah, good, because you actually scared me for a second :D
I was frantically starting to google around whether my understanding of anything and everything is incomplete!

tests/abi/abi.c
Cvelth: [Done](https://github.com/revng/revng-qa/pull/4/commits/d4d3357bc16ea9b5813941ac8de36acd2afb65ed).

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+#include "platform_specific/x86_64.h"
+#elif __x86__
+#include "platform_specific/x86.h"
+#else
+_Static_assert(0, "unsupported platform architecture");
+#endif
+
+/* LFSR generator */
+
+static uint64_t LFSR = 0xdc984ef2647af904ull;
```
aleclearmind: For C files we usually use snake_case, all lower case, see other C files in revng-qa.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+#include "for_each.h"
+#include "functions.h"
+
+#ifdef __x86_64__
+#include "platform_specific/x86_64.h"
+#elif __x86__
+#include "platform_specific/x86.h"
+#else
+_Static_assert(0, "unsupported platform architecture");
+#endif
```
aleclearmind: Consider using `TARGET_${configuration}`, a macro defined in `compiled.cmake`.

tests/abi/platform_specific/x86_64.h
```diff
@@ -0,0 +1,71 @@
...
+#define REG_ "%%"
+#endif
+
+#ifndef LOC_
+#define LOC_ "%0"
+#endif
+
+#define REGISTER_LIST rax, rbx, rcx, rdx, rsi, rdi,        \
+                      r8, r9, r10, r11, r12, r13, r14, r15
+#define REGISTER_C_TYPE uint64_t
```
aleclearmind: This can be a `typedef`.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+
+#define GET_1ST_HELPER(THIS_ONE, ...) THIS_ONE
+#define GET_1ST(THIS_ONE, ...) GET_1ST_HELPER(THIS_ONE, __VA_ARGS__)
+#define FIRST_REGISTER GET_1ST(REGISTER_LIST)
+
+#define STACK_VALUE_COUNT 32
+#define REGISTER_COUNT COUNT_ARGUMENTS(REGISTER_LIST)
+#define GENERATED_COUNT REGISTER_COUNT + STACK_VALUE_COUNT
+
+void generate(REGISTER_C_TYPE *Generated) {
```
aleclearmind: * [ ] Can you add a `count` argument instead of hardcoding `GENERATED_COUNT`? Alternatively, make a `struct CPUState` containing an array with the right number of elements.
* [ ] Maybe `generate_random` is a more appropriate name.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+  do {                                                    \
+    printf("    - Name: \"%s\"\n      Value: 0x%016lx\n", \
+           STRINGIFY(REGISTER),                           \
+           Generated[RegisterIndex--]);                   \
+  } while (0);
+  {
+    uint32_t RegisterIndex = REGISTER_COUNT - 1;
+    FOR_EACH(DUMP_REGISTER, REGISTER_LIST);
+  }
+#undef DUMP_REGISTER
```
aleclearmind: Unless you see a benefit in using macros so aggresively, I suggest to convert `REGISTER_LIST` to a `const char *registers_list[] = { "rax", "rbx", ... };` and drop the related macros.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+    /* Restore the stack */                                                  \
+    RESTORE_STACK(STACK_VALUE_COUNT)                                         \
+    /* Restore all the registers */                                          \
+    REVERSE_FOR_EACH(POP_REGISTER, REGISTER_LIST)                            \
+    :: "X"(Generated)                                                        \
+    : COMMA_SEPARATED_FOR_EACH(STRINGIFY, REGISTER_LIST), "memory"           \
+  );                                                                         \
+                                                                             \
+  free(Generated);                                                           \
+}
```
aleclearmind: This does not need to be a macro, you can pass in a `void *` as argument and call that.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+
+#ifdef __APPLE__
+#define FUNCTION_NAME_PREFIX _
+#else
+#define FUNCTION_NAME_PREFIX
+#endif
+
+#ifdef REG_
+#undef REG_
+#endif
```
aleclearmind: Why?

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+  STRINGIFY(FUNCTION_NAME_PREFIX) "restoreTheState:\n"                     \
+  POP_REGISTER(rbp)                                                        \
+  RESTORE_STACK(STACK_VALUE_COUNT)                                         \
+  REVERSE_FOR_EACH(POP_REGISTER, REGISTER_LIST)                            \
+  RESTORE_STACK(2)                                                         \
+  /* FREE_ON_HEAP(FRAME_POINTER) // A known memory leak. TODO: fix. */     \
+  PUSH_REGISTER(rbp)                                                       \
+  RETURN_FROM_FUNCTION                                                     \
+)
+NAKED_FUNCTIONS;
```
aleclearmind: * [ ] This piece of code is quite obscure. Document what's going on. What are `getTheValue` and `restoreTheState`?
* [ ] Why `#define X` and use it right away? Can't you just drop the macro and use its content right away?

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+#undef LOC_
+#endif
+#define LOC_ "%" STRINGIFY(FRAME_POINTER)
+
+/* clang-format off */
+#define NAKED_FUNCTIONS __asm__ (                                          \
+  /* Define the first function */                                          \
+  STRINGIFY(FUNCTION_NAME_PREFIX) "getTheValue:\n"                         \
+  UPDATE_FRAME_POINTER                                                     \
+  ALLOCATE_ON_HEAP(FRAME_POINTER, (GENERATED_COUNT) * (REGISTER_SIZE))     \
```
aleclearmind: Why do we need to use the heap?

tests/abi/functions.h
```diff
@@ -0,0 +1,273 @@
...
+}
+
+/* clang-format off */
+#define ARGUMENT_FUNCTION_LIST functionWithLotsOfArguments,  \
+                               multitypeArgumentFunction,    \
+                               functionForRandomnessTesting, \
+                               single16bitArgument,          \
+                               single32bitArgument,          \
+                               single64bitArgument,          \
+                               single128bitArgument
```
aleclearmind: Let's try to keep macro magic as local as possible.
```c
typedef struct {
  const char *name;
  void *function;
} test_function_t;

#define ENTRY(name) { #name, name }

test_function_t test_functions[] = {
  ENTRY(functionWithLotsOfArguments),
};

#undef ENTRY
```
And then you can iterate on `test_functions` without macro magic.

tests/abi/functions.h
```diff
@@ -0,0 +1,273 @@
...
+  restoreTheState();
+}
+
+/* clang-format off */
+#define RETURN_VALUE_FUNCTION_LIST single8bitReturnValue,  \
+                                   single16bitReturnValue, \
+                                   single32bitReturnValue, \
+                                   single64bitReturnValue, \
+                                   single128bitReturnValue
+/* clang-format on */
```
aleclearmind: See suggestion for `ARGUMENT_FUNCTION_LIST`

tests/abi/functions.h
```diff
@@ -0,0 +1,273 @@
...
+  PRINT_ARGUMENT_IMPL(uint64_t, "0x%016llx", ARGUMENT_NAME)
+#define PRINT_UINT128_T(ARGUMENT_NAME)                    \
+  PRINT_ARGUMENT_IMPL(__uint128_t,                        \
+                      "0x%016llx%016llx",                 \
+                      (uint64_t) ((ARGUMENT_NAME) >> 64), \
+                      (uint64_t) ARGUMENT_NAME)
+
+/*                                 */
+/* Argument passing test functions */
+/*                                 */
```
aleclearmind:
```suggestion
/*
 * Argument passing test functions
 */
```

tests/abi/functions.h
```diff
@@ -0,0 +1,273 @@
...
+  PRINT_UINT8_T(_12);
+  PRINT_UINT8_T(_13);
+  PRINT_UINT8_T(_14);
+  PRINT_UINT8_T(_15);
+  PRINT_UINT8_T(_16);
+  puts("  ReturnValues:");
+  puts("");
+}
+
+void __attribute__((noinline, weak)) ABIDEF single16bitArgument(uint16_t _00) {
```
aleclearmind: Maybe create a macro for attribute sets you often use.

aleclearmind (changes requested):

tests/abi/abi.c
Cvelth: 1. The `GENERATED_COUNT` is also extensively used at the point which the numbers are generated for. That's why I went with the macro, and I can't afford to spend another register to hold the count at the point where it's needed later down the line, so it has to be defined at compile time.
2. It was a struct initially, I don't remember exactly what ticked me off to go to the inferior "pointer-to-array" approach, but something about the assembly it was generating was messing up the stuff later down the line.
3. I agree about the name.

tests/abi/abi.c
Cvelth: Wait, I'm pretty sure those are only used at compile time, so run-time array doesn't cut it :(
Even more so considering I'm iterating on this list at runtime.

The only possible thing I can see us being able to improve here is going to a macro list of strings, e.g.
```cpp
#define REGISTER_LIST "rax", "rbx", "rcx", [...]
```
but I'd rather use an extra `STRINGIFY` than deal with potential complications of a misplaced `"`, etc (I got burned on that stone once, never again (if I can help it)).
As a bonus `STRINGIFY` is a lot more explicit about what's actually going in the mess down below.

tests/abi/abi.c
Cvelth: It's A LOT simpler to "call" a compile-time function from the inline assembly over the runtime pointer. At least as far as my search went.

I didn't want to imitate templates and spawn a bunch of similar function for each test type as well, but the other option seems even worse.

tests/abi/abi.c
Cvelth: It's a really dirty workaround I had to do because gcc's inline assembly sucks! When the capture-like thing is used, the registers have to use `%%` prefix, when it is not - `%`! And I have to have the capture thing for the assembly inside a c function. And I'm not allowed to have one in the "free" assembly block later down. So it's a way insert the whatever prefix is needed at a time in the right place.

tests/abi/abi.c
Cvelth: 1. Yes, right, I'm sorry. That's my bad. When I managed to get this ~~mess~~ piece of code working, I felt really tired so I decided to do it the next morning. And then forgot and moved on to some other stuff. I'll look into documenting this.
2. That's a trick to avoid our convention checker complaining :) It really doesn't like the first line ending in `(`.

tests/abi/abi.c
Cvelth: Yes, because we're feeling the stack with those values after this. And it's really hard to keep track of where the values come from when the stack changes all the time. When using the heap, those values get generated and just stay at one single point. Only the pointer needs to be kept track of. And that pointer never changes (until `free` is called on it).

tests/abi/functions.h
Cvelth: An `IterationOn` function needs to be generated for each of those, so I wasn't able to do anything about that. Let's solve the [relevant discussion](https://github.com/revng/revng-qa/pull/4#discussion_r763003454) first - that should give an answer to this problem as well.

tests/abi/abi.c
Cvelth: P. S. For where `getTheValue` and `restoreTheState` functions are used you can take a look [here](https://github.com/revng/revng-qa/pull/4/files#diff-69068270dabcdde6ac2fe63aa22ce2d4972973a1b5eee6fae36d18f2cc27c1abR224) and [here](https://github.com/revng/revng-qa/pull/4/files#diff-69068270dabcdde6ac2fe63aa22ce2d4972973a1b5eee6fae36d18f2cc27c1abR231)

P. P. S. I know that technically calling a casted function pointer is UB, but considering that the values of every register and a bunch of stack is literally filled in with predetermined random stuff - that's the least of our problems there :D

tests/abi/abi.c
aleclearmind: > are only used at compile time

What do you mean by that?
```c
const char *registers_list[] = { "rax", "rbx", ... };

void dump_registers() {
  const size_t register_count = elementsof(registers_list);
  uint32_t RegisterIndex = register_count - 1;
  for (unsigned I = 0; I < register_count; ++I) {
     printf("    - Name: \"%s\"\n      Value: 0x%016lx\n",
           registers_list[I],
           Generated[RegisterIndex--]);
  }
}
```

Cvelth: I'm sorry, they are **mostly** used at compile time.
For example
```cpp
/* Save all the registers */
FOR_EACH(PUSH_REGISTER, REGISTER_LIST)
```
Same with `DUMP_REGISTER` - it's used within an unrolled loop to synthesize a compile-time (as far as I understand) string used with `__asm__`.

Cvelth: And compile time loops are important since we can't exactly tell the compiler to skip one register to use for looping, etc. I feel a lot safer if every single loop is already unrolled before the compiler even touches it - even if we went the route of splitting the `__asm__` declaration into multiple smaller ones.

tests/abi/abi.c
aleclearmind: Ah because you don't want to waste a register, uh?
In CISC architecture we can probably get away with something, but in RISC.
Gosh... This is tricky to get right.

Please add a comment stating why this needs to be a distinct function.

Cvelth: Believe me, I would have never chosen to ~~suffer~~ create macro hell if I saw another viable alternative.

tests/abi/abi.c
aleclearmind: I see, I hope this ASM strings concatenation magic turns out to be suitable for RISC architectures. Probably it does.

What do you think of generating `const char *registers_list[]` from `REGISTER_LIST` just for the sake avoiding one usage of `FOR_EACH(DUMP_REGISTER, REGISTER_LIST);`?

tests/abi/abi.c
aleclearmind: Can you see if you can solve the problem with an extra macro argument for macros using `REG_`?
Or maybe have two distinct macros.

tests/abi/abi.c
Cvelth: I'm not opposed, but it's more of a consistency issue at that point. If we're already using this uber-ugly (but more optimized and unavoidable elsewhere) solution, we might as well keep this one avoidable point the same.

I'm fine either way, though (the way we dump stuff is definitely the least of my worries about this thing ;)).

tests/abi/abi.c
Cvelth: Problem is that everything is loops on top of loops. And this was a super-late addition to the fray, so I decided against rewriting everything from scratch just to accommodate for this (this was already taking way longer than it should have at that point, even more so now) - meaning to add an extra argument to most of the stuff and to rework loops to propagate it.

And "two distinct macros" would probably mean two distinct `x86_64.h`-like files at that point. Since stuff depends on other stuff.

tests/abi/abi.c
```diff
@@ -0,0 +1,175 @@
...
+  FOR_EACH(PUSH_REGISTER, REGISTER_LIST)                                   \
+  REPEAT(PUSH_CONSTANT, STACK_VALUE_COUNT, REGISTER_COUNT, FIRST_REGISTER) \
+  INDEXED_FOR_EACH(FILL_REGISTER, REGISTER_LIST)                           \
+                                                                           \
+  FIX_RETURN_ADDRESS((GENERATED_COUNT + 1) * (REGISTER_SIZE))              \
+  RETURN_FROM_FUNCTION                                                     \
+                                                                           \
+  /* Define the second function */                                         \
+  STRINGIFY(FUNCTION_NAME_PREFIX) "restoreTheState:\n"                     \
+  POP_REGISTER(rbp)                                                        \
```
aleclearmind: This is not portable

Cvelth: Whoops, right, my bad, there's already macro for that (see [here](https://github.com/revng/revng-qa/pull/4/files#diff-86d497e9de7e66dcdab3f18d737c7e11b83568d711458596246551f26e75d5e6R135)). That would have been easily caught as soon as I added one more supported architecture. But thanks for pointing it out, I'll fix it.

tests/abi/abi.c
aleclearmind: 1. What are they doing though? Seems like a sort of `setjmp`/`longjmp`, you want to use those, maybe?
2. Add a comment or whatever, let's not add one single more macro then necessary `:)`

Cvelth: 1. They are called from a c-function. `getTheValue` one sets the state up (fills registers and stack). The [test function](https://github.com/revng/revng-qa/pull/4/files#diff-69068270dabcdde6ac2fe63aa22ce2d4972973a1b5eee6fae36d18f2cc27c1abR225) calls it and pretends to receive some value - which is then `printf`'ed. After which it calls `restoreTheState` which makes sure all the register values and the stack are restored to the way that would have been if none of this happened.
2. 👍🏻

tests/abi/abi.c
aleclearmind: I agree, but let's push to the minimum the macros. They spread like XML.

tests/abi/abi.c
aleclearmind: You could:
1. `x = malloc(1000)`
2. `setjmp`
3. Pivot the stack to the `x` area
4. Call the function
5. Print the return value
6. `longjmp`
7. `free(x)`

This way you won't need `restoreTheState` nor the part of `getValue` that saves it!

tests/abi/abi.c
aleclearmind: AFAIU we could use the capture-like `__asm__` except for `getTheValue` (and `restoreState`).
Let's see if we can get rid/inline them and switch to only use `__asm__` with the capture-like thing.

tests/abi/abi.c
Cvelth: Wait, right, how have this never came up when I was trying to find a cleaner way to do it. This actually makes a whole lot of sense. The only question is how portable is `setjmp/longjmp` pair? Since it seems to be part of linux kernel.

tests/abi/abi.c
Cvelth: that would be the dream come true!

tests/abi/abi.c
aleclearmind: `setjmp` is just user space.
`sigsetjmp` does some signal stuff which might be non-portable, but I see no reason why `mingw` wouldn't implement `setjmp`.

Also, instead of `malloc`/`free`, consider using a global variable, simpler.

tests/abi/abi.c
Cvelth: Should be done. Please let me know if maybe I missed something.

tests/abi/abi.c
Cvelth: Done.

tests/abi/platform_specific/x86_64.h
Cvelth: Done.

tests/abi/abi.c
Cvelth: 1 & 2. Irrelevant since it became a global variable.
3. I chose to rename the whole thing into `random_state`, `regenerate_state` and `print_current_state`, what do you think?

tests/abi/abi.c
Cvelth: Should be done. Please let me know what you think.

tests/abi/abi.c
Cvelth: Ok, I have an idea, maybe we can avoid this extra macro by using a global variable-trick or something.

The question is which evil do you hate more? xD

tests/abi/abi.c
Cvelth: Done, done, and done! And I couldn't be happier about that!

tests/abi/abi.c
Cvelth: Done.

tests/abi/functions.h
Cvelth: Done.

tests/abi/abi.c
Cvelth: Gone.

tests/abi/abi.c
aleclearmind: The issue is that in RISC architectures you call a function pointer in a global variable without materializing it in a register.
Either we define a *scratch* register for each architecture, or we need to stick with something like this, I'm afraid.

Cvelth: That's what I was afraid of, I guess this thread can be closed then.

tests/abi/functions.h
Cvelth: Done.

tests/abi/abi.c
Cvelth: Heap usage is gone (yay)

tests/abi/functions.h
Cvelth: So, to recap this, using runtime function pointers is painful, I'd rather let linker take care of the converting function names to addresses - hence the compilation time loops.

tests/abi/runtime_abi_analysis.c
```diff
@@ -0,0 +1,156 @@
...
+    randomized_state[counter] = get_next_lfsr();
+}
+void print_current_state(void) {
+  /* Dump Stack */
+  puts("- Stack:");
+  for (uint32_t index = REGISTER_COUNT; index < GENERATED_COUNT; ++index) {
+    printf("    - Offset: 0x%02x\n      Value: 0x",
+           (index - REGISTER_COUNT) * REGISTER_SIZE);
+    PRINT_VALUE(uint64_t, randomized_state[index]);
+    puts("");
```
aleclearmind: What's the advantage of printing it at 64-bit shots compared to a by array?

tests/abi/functions.h
```diff
@@ -0,0 +1,309 @@
...
+       "  IsLittleEndian: no");
+#endif
+
+#define UNROLL_ARGUMENT(LHS, RHS) DO_NOTHING(LHS, argument_##RHS)
+
+size_t runtime_endianness_check() {
+  size_t i = 1;
+  return !*((char *) &i);
+}
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
```
aleclearmind:
```suggestion

#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
```

tests/abi/functions.h
```diff
@@ -0,0 +1,309 @@
...
+  do {                                   \
+    typedef union {                      \
+      TYPE v;                            \
+      uint8_t a[sizeof(TYPE)];           \
+    } printing_helper;                   \
+                                         \
+    printing_helper helper;              \
+    helper.v = VALUE;                    \
+                                         \
+    PRINT_BYTES(helper.a, sizeof(TYPE)); \
```
aleclearmind: Why handle the endianess by hand when you could just `printf("%x")`?

tests/abi/functions.h
```diff
@@ -0,0 +1,309 @@
...
+  } while (0);
+
+#define PRINT_VARIABLE(TYPE, VALUE)                                        \
+  do {                                                                     \
+    printf("    - Type: " #TYPE "\n      Address: 0x%0x\n      Value: 0x", \
+           &VALUE);                                                        \
```
aleclearmind: Why print the `Address`?

tests/abi/functions.h
```diff
@@ -0,0 +1,309 @@
...
+
+static_assert(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+                || __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__,
+              "Unsupported endianness (pdp? something crazy? Memory "
+              "corruption?)");
+
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define PRINT_FUNCTION_HEADER(FUNCTION_NAME)            \
+  puts("  Function: \"" STRINGIFY(FUNCTION_NAME) "\"\n" \
+       "  IsLittleEndian: yes");
```
aleclearmind: Why print endianess really relevant?

tests/abi/reference_binary.c
```diff
@@ -0,0 +1,25 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+
+#include <setjmp.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "functions.h"
```
aleclearmind: Headers that have "side effects" (e.g., that you cannot include more than once, or define non-`static` functions) usually have the `.inc` extension.

tests/abi/reference_binary.c
```diff
@@ -0,0 +1,25 @@
...
+int main(void) {
+
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+  assert(!runtime_endianness_check());
+#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+  assert(runtime_endianness_check());
+#endif
+
+  return 0;
+};
```
aleclearmind: Can we merge `reference_binary.c` and `runtime_abi_analysis.c`?

Also, in the `abi/` directory we need a `README.md` documenting what's going on.
Feel free to move everything you deem relevant here from the `README.md` in revng.

tests/abi/ABITests.cmake
```diff
@@ -0,0 +1,21 @@
...
+
+macro(register_abi_artifacts ABI ARCHITECTURE)
+
+register_artifact_category("reference_abi_binary_${ABI}" "ON")
+register_artifact("reference_abi_binary_${ABI}" "reference_binary" "${ARCHITECTURE}" "reference_binary.c")
+register_artifact_run("reference_abi_binary_${ABI}" "reference_binary" "default" "nope")
+
+register_artifact_category("runtime_abi_analysis_${ABI}" "ON")
+register_artifact("runtime_abi_analysis_${ABI}" "analyzed_binary" "${ARCHITECTURE}" "runtime_abi_analysis.c")
+register_artifact_run("runtime_abi_analysis_${ABI}" "analyzed_binary" "default" "nope")
```
aleclearmind: Not sure *analyzed* correctly captures what this binary does.
Maybe something more about "introspection"? But let's first decide whether it makes sense to merge `reference_binary.c` and `runtime_abi_analysis.c` or not.

aleclearmind (changes requested):

tests/abi/runtime_abi_analysis.c
Cvelth: Initially I was intending on doing more with the stack - but that turned out impractical. So yeah, as things stand, just an array of "numbers" should work just as well.

tests/abi/functions.h
Cvelth: To support objects of arbitrary size without worrying about splitting them into register-sized burst and/or system-dependent endianness conversion methods.

tests/abi/functions.h
Cvelth: Again, I had deeper intent for it, but it lost its meaning when the "dump" evolved. I'm just hesitant to remove it just in case I find some use for it.

tests/abi/functions.h
Cvelth: This helps making the artifact self-sufficient. As in, the parsing side don't need to be aware of the system endianness - it can just get it from here.

tests/abi/reference_binary.c
Cvelth: Sure.

tests/abi/reference_binary.c
Cvelth: I don't want the extra functions `analysis` introduces to be caught in the cross-fire when parsing the binary. The reference is like it sounds - only the functions we're interested in while `analysis` also contains the way for them to dump stuff and so on.

tests/abi/ABITests.cmake
Cvelth: Totally agree - this name sucks but I never came op with anything better.

tests/abi/runtime_abi_analysis.c
aleclearmind: I think it'd be simpler.
But I don't want to make this longer than necessary. Maybe put a TODO. Also, please document in `README.md` where offset `0` is. From a callee's perspective, is it the integer right after the saved return address?

tests/abi/functions.h
aleclearmind: OK, but here, correct me if I'm wrong, you're using this function to print the value contained in a register, which is either 32-bit or 64-bits. `printf` would have the same effect, with less logic. No need to perform any conversion.

tests/abi/functions.h
aleclearmind: The testing infrastructure is quite complex, let's strive to shrink it as much as we can.
Unless you think this is something hard to re-introduce again in case of need, but it doesn't seem to be the case.

tests/abi/runtime_abi_analysis.c
Cvelth: So it's pretty much
```
push rcx // offset 16
push rbx // offset 8
push rax // offset 0
call my_function
```
There's no easy way to find out whether there is an address on the stack, shadow space for something or anything else - so I didn't bother. That represents `push`es right before the call instruction.

tests/abi/reference_binary.c
aleclearmind: Uhm, what about:

* `reference_binary.c` -> `abi-test-functions-library`
* `runtime_abi_analysis.c` -> `describe-abi-test-functions`

tests/abi/ABITests.cmake
aleclearmind: Let's keep discussing in the other one.

tests/abi/functions.h
Cvelth: I have exactly one printing "macro" which is used for everything. Be it a register or a 80-byte struct. And it's this `PRINT_BYTES` thing.

tests/abi/reference_binary.c
Cvelth: I like the way this sounds!

tests/abi/runtime_abi_analysis.c
aleclearmind: Uhm, if that's exactly the code you have, you know 100% sure that the offset 0 integer is going to be right after the saved return address pushed by the `call` instruction.
If this is the case, what I suggested before should hold, right? Doc doc doc.

Cvelth: Not every architecture pushes the address. Also they might theoretically push something else on top of the return address - that's why I'm trying to make as few assumptions as I can here.

tests/abi/functions.h
aleclearmind: Mmmmh, why would you print an 80-byte `struct` in an endianess aware way? It's not what you want to do.
Endianess is a concept that only makes sense for integers that are loaded/stored from memory in a single shot. Which typically means 32- or 64-bit integers.

You should not print a 80-bytes struct composed by multiple integers in a direction or the other depending on the system endianess. Or am I misunderstanding what you're saying?

Cvelth: It's probably less about you misunderstanding what I'm saying and more of me not having the complete understanding of what's going on.

Here are the things the current implementation is built to guarantee:
1. The values of all the registers and the stack "cell"s are printed the same way lldb prints them.
2. All the values of arguments/return values are printed using this endianness-aware macro on top of `printf`.
3. This comparison works (pseudocode): 
    ```cpp
    uint64_t Parsed;
    ArgumentValue.substring(CurrentChunkOffset, 16).getAsInteger(16, Parsed);
    return Parsed == Expected_value_of_this_register_or_stack_cell;
    ```
Maybe the semantics are not as they are expected, but a bunch of trial and error led to this way of comparing the values.

tests/abi/runtime_abi_analysis.c
aleclearmind: I see two options: either `call` pushes the return address on the stack or nothing is pushed.
In both cases we can say that offset `0` corresponds to the value of the stack pointer *right before* the `call` instruction. Is this right?

> I'm trying to make as few assumptions as I can

Making as few assumptions as possible is important, but it cannot become a reason not to document and provide sane expectation to the next dev that will look at this code `:)`.

Cvelth: Yeah, this makes sense. Sorry I confused you.

The point is that I cannot control the `call` instruction itself ~~and so it scares me~~.

aleclearmind: Aaah. I missed that part.
I think that if there's anything in there the test is f-'d.
I think we either find a way to `assert` things are the way we expect, or we should just assume it and document things accordingly.

Cvelth: Yeah, pretty much. And that's what makes me so sad about the situation.

aleclearmind: Wait, why do you say that you cannot control the call instruction?

I see [the following](https://github.com/revng/revng-qa/pull/4/files#diff-294dd971aa761265be5b8dcc15dfaad6bee2d5ddb4413d015a1157e61767d66cR99):

```

    __asm__ (                                                                 \
      /* Popularize the stack with random data */                             \
      REPEAT(PUSH_CONSTANT, STACK_VALUE_COUNT, GENERATED_COUNT, TMP_REGISTER) \
      /* Popularize all the registers with random data */                     \
      INDEXED_FOR_EACH(FILL_REGISTER, REGISTER_LIST)                          \
                                                                              \
      /* Call the function */                                                 \
      CALL_A_FUNCTION(FUNCTION_NAME)                                          \
                                                                              \
      /* Restore the stack */                                                 \
      RESTORE_STACK(STACK_VALUE_COUNT)                                        \
    );                                                                        \
```

We control the call!

Cvelth: We control what happens BEFORE the `call`. And what happens AFTER the `call`. What I'm scared of is what the call instruction itself does. Sure, it's pretty simple on most common architectures. But some exotic ones that do some weird stuff probably exist.

aleclearmind: I like to make statements that can be either true or false, no matter ABI, architecture or whatever.
Here's my statement:
> Offset 0 corresponds to the value of the stack pointer right before the call instruction.

True or false? If it's true, I think it's an important piece of information that we should put down as a definition of what the `Stack` data in the YAML is.

Cvelth: I'd phrase it like

> Offset `0` indicates the value corresponding to the last unpopped register-sized value on the stack right before the call instruction. Offset `8/4/RegisterSize` - second to last unpopped value. Offset 2*`8/4/RegisterSize` - third and so on.

Does that work?

tests/abi/functions.h
Cvelth: This is gone.

tests/abi/functions.h
Cvelth: This is gone (yay!!)

tests/abi/functions.h
Cvelth: Go-o-o-one!

tests/abi/functions.h
Cvelth: `Address` is gone for now (and hopefully for good).

tests/abi/runtime_abi_analysis.c
aleclearmind: Not really, that's very long and convoluted.
Can you make an example that falsifies the one above?

Cvelth: They are not mutually exclusive. I just extended it with basic information of what other offsets mean.

aleclearmind: If you say "unpopped" you're forcing the user to know that something has been pushed, which is an implementation detail to me. There could be an implementation with no-push and no-pops.
If you want, you can add "assuming the stack grows towards low addresses, ...". But other than that I'd try to explain what that data is in a simple way, leaving out implementation details.

Cvelth: Sure, I'm fine with that.

tests/abi/reference_binary.c
Cvelth: Done: https://github.com/revng/revng-qa/pull/4/commits/b375463f0ba22f1fa4cd2f8240d50300872e5d22.

tests/abi/reference_binary.c
Cvelth: Done: https://github.com/revng/revng-qa/pull/4/commits/614d116fd2d54bdcf22308287880bffe89dde361.
(also there's a revng#148 commit for this)

PR #7, opened 2022-04-06
Reimplement the abi artifact generator on top of jinja2

Cvelth:


tests/abi/include/common.h
```diff
@@ -0,0 +1,64 @@
...
+      printf("0x%.2hhx, ", (HELPER)->a[i]);                \
+    printf("0x%.2hhx ]\n", (HELPER)->a[sizeof(TYPE) - 1]); \
+  } while (0)
+
+#define PRINT(TYPE, POINTER)                            \
+  do {                                                  \
+    printf("      - Type: " #TYPE "\n        Bytes: "); \
+    MAKE_PRINT_HELPER(TYPE, POINTER, local_helper);     \
+    PRINT_BYTES(TYPE, local_helper);                    \
+  } while (0)
```
aleclearmind, line 73: Can we get rid of these too?

scripts/abi_artifact_generator/generate.py
```diff
@@ -0,0 +1,202 @@
...
+    type=str,
+    help="Specifies the directory where to output the results to",
+)
+
+arguments = parser.parse_args()
+
+#
+# Load configs
+#
+import yaml
```
aleclearmind, line 9: Accumulate imports on top

scripts/abi_artifact_generator/generate.py
```diff
@@ -0,0 +1,202 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+#
+# Accept arguments (only paths for now)
+#
+import argparse
+
+parser = argparse.ArgumentParser(
```
aleclearmind: Push all the top-level code in a function called `main`.

The typical pattern is

```python

def main():
    return 0

if __name__ == "__main__":
    sys.exit(main())
```

Most importantly let's not interleave module-level code with function definitions.

aleclearmind (changes requested):

scripts/abi_artifact_generator/generate.py
Cvelth: Done: https://github.com/revng/revng-qa/pull/7/commits/e945e386f6110b972949e3e5a21fd047cd0da6bb.

scripts/abi_artifact_generator/generate.py
Cvelth: Something like this?
https://github.com/revng/revng-qa/pull/7/commits/3ed002e5d4f07347161c7a4667cc95aec981c908

tests/abi/include/common.h
Cvelth: Closing this based on the mattermost discussion.

tests/abi/ABITests.cmake
```diff
@@ -2,20 +2,40 @@
...
 
+# Ensure the generator script is run only inside the `revng-qa` project
+if("${CMAKE_PROJECT_NAME}" STREQUAL "revng-qa")
+  execute_process(
+    COMMAND
+      "python3"
+      "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/generate.py"
+      "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/templates/"
+      "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/config/"
+      "${CMAKE_SOURCE_DIR}/tests/abi/generated/")
```
aleclearmind: * `execute_process` runs at configure-time. I think we want to run this at build time with `add_custom_command`/`add_custom_target`.
* Generate the source file in the build directory, not the source directory.

Cvelth: 1. Yeah, I'm aware, this is the side-effect of trying to deal with the complications around the second problem - so if we deal with it - we get this for free.

2. Now to the core issue. I did fix `register_artifact` to actually support multiple source files. But due to how it's structured, having those in wildly different directories still doesn't work nicely. The problem is that we'd need to provide the "common" path piece and then subpaths from it. Except if one of them is in `${CMAKE_SOURCE_DIR}` and the other one is in `${CMAKE_BUILD_DIR}` - we should not try to lump them together - there's no guaranties about how they relate to one another. I'm afraid that fixing this would require quite a substantial rework of `register_artifact` macro and also modifications to how its users provide source paths.

P. S. I'm sorry I forgot to bring this up like I intended to do when I decided to temporarily generate files in the source directory and "deal with it later".

tests/abi/ABITests.cmake
```diff
@@ -2,35 +2,34 @@
...
     COMMAND
       "python3"
       "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/generate.py"
       "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/templates/"
       "${CMAKE_SOURCE_DIR}/scripts/abi_artifact_generator/config/"
-      "${CMAKE_SOURCE_DIR}/tests/abi/generated/")
+      "${CMAKE_BINARY_DIR}/include/revng-qa/generated/")
+else()
+  add_custom_target(generated_abi_artifact_sources)
 endif()
```
Cvelth: Having to only define the target within `revng-qa` feels really bad - but since `revng` also includes this file as is (due to how the relation between the two projects is structured) I don't see a way around it. After all I don't want to duplicate the generation script, templates and configs into both `${CMAKE_SOURCE_DIR}` directories.

macros.cmake
```diff
@@ -231,7 +231,7 @@ macro(register_derived_artifact FROM_ARTIFACTS NAME SUFFIX TYPE)
             add_custom_target(
               compile-${NAME}_${ARTIFACT_CATEGORY}__${ARTIFACT}__${CONFIGURATION}
               ALL
-              DEPENDS ${OUTPUT} ${DEPEND_ON})
+              DEPENDS ${OUTPUT} ${DEPEND_ON} generated_abi_artifact_sources)
```
Cvelth, line 234: I hate adding the dependency in such a clumsy manner. It works - sure, but there HAS TO be a better way. This also might be helpful in solving the other issue as well.

Do you have any ideas, @aleclearmind?

tests/abi/ABITests.cmake
aleclearmind: One day this repo will disappear, for now it's OK.

macros.cmake
aleclearmind: You could create a variable to define the variable and define it to an empty string if not revng-qa.

Cvelth: Yeah, I see your point.

But that's not the main issue I'm concerned about. Having this dependency outside of `revng-qa` is not really a problem since creating an "empty" target is pretty much free.

What I hate is the idea that this forces totally unrelated targets (all of them, to be exact) to depend on a single script they don't even need to depend on - de facto creating a bottleneck - no target can start building until the script is over.

aleclearmind: `revng-qa` currently takes no time to build. Let's not optimize it if it's not broken.
Also, I'm trying to re-engineer all of this from scratch, so no need to make it good.

Cvelth: If so I'd rather "everything clearly depends on the generator" over "everything depends on a variable to define the variable" just to hide it away (without fundamentally changing anything).

But it's not a strong opinion, either way works.

What's your stance? 

And also, should I squash these commits right away or do you intend to review them separately?

aleclearmind: I'm OK with either options. Important part is that we don't use `execute_process`/generate stuff in the build directory.

Squash squash squash

tests/abi/ABITests.cmake
Cvelth: Done.

PR #9, opened 2022-06-28
Introduce import-pe tests

dwarfyresearch:
This adds sources for ```revng model import``` of PE/COFF.

### DIFF
--- a/share/revng/test/configuration/revng-qa/for-import-pe.yml
+++ b/share/revng/test/configuration/revng-qa/for-import-pe.yml
@@ -1,0 +1,37 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+tags:
+- name: windows
+  variables:
+    CFLAGS:
+      - /DEBUG:FULL
+      - /Z7
+      - /EHsc
+      - /WX
+      - /nologo
+    LDFLAGS:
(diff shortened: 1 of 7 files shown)


share/revng/test/configuration/revng-qa/for-import-pe.yml
```diff
@@ -0,0 +1,20 @@
...
+    WINCFLAGS:
+      - /DEBUG:FASTLINK
+      - /Zi
+      - /EHsc
+sources:
+- tags: [win-flags]
+  members:
+  - tests/model/import/pe/main.c
+commands:
+- type: revng-qa.win-compiled
```
aleclearmind: The fact that you're not using `revng-qa.compiled` is preventing this from going through the whole pipeline (down to decompilation), is this intended?
I think you can have multiple commands with the same `type`.

If we do this, we could have something like

```yaml
commands:
- type: revng-qa.compiled
  from:
  - type: source
    filter: windows
  command: x86_64-winsdk-vc16-cl ...
- type: revng-qa.compiled
  from:
  - type: source
    filter: linux
  command: ${TRIPLE}gcc ...
```

We could also reuse the variable `CFLAGS` instead of `WINCFLAGS` which sounds "exceptional".

What do you think?

PR #12, opened 2022-10-06
Extend the abi testing suite

Cvelth:
This is a sister PR of [`revng#232`](https://github.com/revng/revng/pull/232).

### DIFF
--- a/scripts/abi_artifact_generator/generate.py
+++ b/scripts/abi_artifact_generator/generate.py
@@ -93,9 +93,19 @@ def setup_stack(architecture, config):
     first_offset = last_offset + config["stack_byte_count"]
 
     result = str("")
+
+    # Some architectures only support pushing registers onto the stack in pairs.
+    # To work around that limitation, this introduces the `fill_two_registers`
+    # option for the templates to utilize: it lets two registers to be loaded
+    # at once, but has the limitation of feeling up twice the space - hence we
+    # need to limit the pushes, which is what the `flip_flop` flag is for:
+    # the registers are only pushed when it's set, which happens on every second
+    # iteration.
+    flip_flop = True
     for offset in range(first_offset, last_offset, -register_size):
         stack_helper_dictionary = {
             "register": architecture["register_list"][0],
+            "second_register": architecture["register_list"][1],
             "fill_a_register": fill_a_register(
                 architecture,
                 "randomized_state",
@@ -103,10 +113,30 @@ def setup_stack(architecture, config):
                 architecture["register_list"][0],
             ),
         }
+        if flip_flop:
+            stack_helper_dictionary["fill_two_registers"] = (
+                fill_a_register(
+                    architecture,
+                    "randomized_state",
+                    offset,
+                    architecture["register_list"][0],
+                )
+                + "\n"
+                + fill_a_register(
+                    architecture,
+                    "randomized_state",
+                    offset - register_size,
+                    architecture["register_list"][1],
+                )
+            )
+        else:
+            stack_helper_dictionary["fill_two_registers"] = ""
         jinja_template = jinja2.Environment().from_string(
             architecture["templates"]["push_to_stack"]
         )
         result = result + jinja_template.render(stack_helper_dictionary) + "\n"
+
+        flip_flop = not flip_flop
 
     return result
 
--- a/scripts/abi_artifact_generator/generate.py
+++ b/scripts/abi_artifact_generator/generate.py
@@ -3,9 +3,10 @@
 #
 
 import argparse
-import jinja2
 import os
 import sys
+
+import jinja2
 import yaml
 
 
@@ -71,9 +72,7 @@ def fill_a_register(architecture, location, offset, register):
         "register": register,
     }
 
-    jinja_template = jinja2.Environment().from_string(
-        architecture["templates"]["fill_a_register"]
-    )
+    jinja_template = jinja2.Environment().from_string(architecture["templates"]["fill_a_register"])
     return jinja_template.render(local_dictionary)
 
 
@@ -113,16 +112,12 @@ def setup_stack(architecture, config):
 
 
 def call_a_function(architecture, name: str):
-    jinja_template = jinja2.Environment().from_string(
-        architecture["templates"]["call_a_function"]
-    )
+    jinja_template = jinja2.Environment().from_string(architecture["templates"]["call_a_function"])
     return jinja_template.render({"function_name": name})
 
 
 def restore_stack(architecture, config):
-    jinja_template = jinja2.Environment().from_string(
-        architecture["templates"]["restore_stack"]
-    )
+    jinja_template = jinja2.Environment().from_string(architecture["templates"]["restore_stack"])
     return jinja_template.render({"stack_size": config["stack_byte_count"]})
 
 
--- a/scripts/abi_artifact_generator/generate.py
+++ b/scripts/abi_artifact_generator/generate.py
@@ -185,7 +185,7 @@ def render_functions(jinja_environment, config, out_dir: str):
         "generation_notice": get_generation_notice(),
         "structs": config["helper_structs"],
         "argument_functions": config["argument_tests"],
-        "return_value_functions": config["small_return_value_tests"],
+        "return_value_functions": config["return_value_tests"],
     }
 
     filename = "functions.h"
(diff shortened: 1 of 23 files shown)


scripts/abi_artifact_generator/generate.py
```diff
@@ -93,21 +93,43 @@ def setup_stack(architecture, config):
     first_offset = last_offset + config["stack_byte_count"]
 
     result = str("")
+    flip_flop = True
```
aleclearmind, line 104: Document what's this for

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -37,17 +39,44 @@ void __attribute__((noinline, weak)) ABIDEF /*( name )*/(
...
   if (setjmp(jump_buffer) == 0) {
-    /*( return_value_type )*/ return_value = ((
-      /*(- return_value_type )*/ ABIDEF(*)(void)) typeless_setup)();
-    puts("    Function: \"/*( name )*/\"");
-    puts("    Arguments:");
-    puts("    ReturnValue:");
+    fflush(stdout);
+
+    /*( return_value_type )*/ return_value;
+    if (CHECK_TYPE(return_value) && CHECK_ALIGNMENT(/*( return_value_type )*/)) {
```
aleclearmind: Obscure: what does `true` mean in this `if`? The names of the macros don't tell me much.

Also: do we really really *really* need macros here? We turned everything to jinja in order to avoid macros.

aleclearmind (changes requested):

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: > what does `true` mean in this `if`?

That the return value does not qualify as a `SPARV`, sadly there is no better way for detecting those, rather than doing the alignment checks + encoding some information about the ABI as a pre-defined macros coming from our testing pipeline.

> do we really really _really_ need macros here? We turned everything to jinja in order to avoid macros.

The jinja generates quite a bit of code, which is a pain to work with / debug as is. Hiding some annoying stuff like this away behind `PRINT(x)` macro makes it somewhat manageable, even if not perfect.

TL;DR: I'd rather work with
```cpp
PRINT(int64_t, argument_0);
PRINT(void *, argument_1);
PRINT(int64_t, argument_2);
PRINT(some_struct, argument_3);
```
inside the generated functions over having this ugly and hard to read stuff repeated for each argument.

In a perfect world this would have been a normal c function, but we can't afford to place a call there, it breaks stuff (since we're already abusing what a compiler is willing to put up with quite a bit).

aleclearmind: I'm not so concerned about `PRINT`. I'm more concerned about `CHECK_*`.

It'd be much better to have:

```
/*{ if is_sptar(return_value_type) }*/
```

If not viable, at very least reduce the macros usage doing something like:

```c
bool IsStruct = ...;
bool NonByteAligned = _Alignof(TYPE) != 1 || sizeof(TYPE) == 1;
bool IsSPTAR = NonByteAligned && ((IsStruct && sizeof(TYPE) < ...) || (!IsStruct && sizeof(TYPE) < ...);
if (IsSPTAR) {
```

Or something along those lines.

I really prefer not to have macros creep in again in here, unless absolutely necessary. It's a slippery slope towards having the worse of both worlds.

Cvelth: Ah, wait, sorry for misleading you (I misunderstood my own code). This is not related to `PRINT` stuff, this does indeed get generated as is.

The reason why it's not a good idea to do it in jinja-level is because it operates on per-architecture basis. As in, it only cares about the assembly being the same. So if multiple ABIs use the same architecture, there's not much reason to generate multiples of the same file.

The big pain-point of this check, is to some how pass in the information about the biggest types allowed in the registers, I mean [this stuff](https://github.com/Cvelth/revng-qa/blob/cf9821fc5228a461ff22e5f83fd853f6521bcdf0/share/revng/test/configuration/revng-qa/for-detect-abi.yml#L29-L30).

The whole point of the check is pretty simple:
* `CHECK_TYPE`: determine if the type is considered a primitive or an aggregate by the compiler (using `_Generic`) and then compare the size against the corresponding constant.
* `CHECK_ALIGNMENT`: do not let any "packed" types get in the first branch.  
By packed here, I'm referring the types for which `gcc` reports alignment of `1` even though their size is bigger. This check is definitely over-simplified, and it limits the available types we can test (it makes it impossible to work with stuff like `struct { char a[4]; };`, because it's a non-`1`-sized struct with legit alignment of `1`, but I'd say it's not too bad of a trade-off, considering how limited our options are here.

With a bit of cleanup and better naming, these macros wouldn't look too bad, imo.

P. S. And also I'll put a bit of an excuse, I kind of expected you to not like the extra macros here, so I just quickly wrote something that "works" without the regard for how it looks to go further without stopping on it with the idea that I'll work it out later, with your feedback. Except I got occupied with other stuff and totally forgot this thing existed 😅.

TL;DR: so, our options:
1. (my preferred one) Cleanup the macros, considering that the underlying logic is pretty simple, they shouldn't look too bad.
2. Make the jinja generate a set of sources per-abi instead of per-architecture. Which will grow pretty confusing pretty fast. Consider how many microsoft ABIs there are, for example. Do we really want a separate c-sources for each of them that looks almost exactly like all the rest, except for a single if (which is already a runtime check, C is very limited in ways the compilation-time logic can be expressed, and we can't move this onto the macro time because of `_Generic`) is some of the generated functions?

Cvelth: I'm ending up with
```cpp
#define IS_SCALAR(VALUE) _Generic(VALUE,             \
  uint8_t: 1, uint16_t: 1, uint32_t: 1, uint64_t: 1, \
  int8_t:  1, int16_t:  1, int32_t:  1, int64_t:  1, \
  default: (__builtin_classify_type(VALUE) == 5 ? 1 : 0))
// `__builtin_classify_type` returns `5` for pointers.

// . . .

    bool IsScalar = IS_SCALAR(return_value);
    size_t SizeLimit = IsScalar ? SCALAR_SIZE_LIMIT : AGGREGATE_SIZE_LIMIT;
    bool WithinSizeLimit = sizeof(return_value) <= SizeLimit * sizeof(void *);
    bool IsPacked = (_Alignof(TYPE) == 1 && sizeof(TYPE) != 1);
    if (WithinSizeLimit && !IsPacked) {
      // use registers.
    } else {
      // use a pointer.
    }
```

What do you think?

Cvelth: Also, I just realized that it wasn't working properly for big types, so it's actually:
```cpp
#ifdef __SIZEOF_INT128__
#define INT128_IS_SCALAR 1
#else
#define INT128_IS_SCALAR 0
#endif

#define IS_SCALAR(VALUE)                                            \
  _Generic(VALUE,                                                   \
           int8_t : 1,                                              \
           uint8_t : 1,                                             \
           int16_t : 1,                                             \
           uint16_t : 1,                                            \
           int32_t : 1,                                             \
           uint32_t : 1,                                            \
           int64_t : 1,                                             \
           uint64_t : 1,                                            \
           INT128_T : INT128_IS_SCALAR,                             \
           UINT128_T : INT128_IS_SCALAR,                            \
           default : (__builtin_classify_type(VALUE) == 5 ? 1 : 0)) \
           // `__builtin_classify_type` returns `5` for pointers.
```

aleclearmind: Works.
`#define` something for `5`.

scripts/abi_artifact_generator/generate.py
Cvelth: Something like this?
```python
# Some architectures only support pushing registers onto the stack in pairs.
# To work around that limitation, this introduces the `fill_two_registers`
# option for the templates to utilize: it lets two registers to be loaded
# at once, but has the limitation of filling up twice the space - hence we
# need to limit the pushes, which is what the `flip_flop` flag is for:
# the registers are only pushed when it's set, which happens on every second
# iteration.
```

aleclearmind: Works

aleclearmind:
Merged in 98798c3. Thanks!

PR #15, opened 2023-02-07
common.sh: Delete temporaries

mrjackv:
When running tests, the `temp` function would be used in a subshell in a similar vein to `$(mktemp)` to create a temporary file/folder. In order to track these files an array was used, however due to the use of a subshell this variable would not be propagated to the parent process. Instead, this commit uses a temporary file to store the paths of the newly created files/folders so that they can be later deleted by the parent process.


aleclearmind:
Merged in 6e555be. Thanks!

PR #16, opened 2023-03-21
Minor improvements for `revng-test-configure` script

Cvelth:
This is a sister PR for https://github.com/revng/revng/pull/274

The similarity is defined as the length of the common chain of the steps taken from the source.
But, when comparing two tuples, the choice made on that characteristic alone is not sufficient for the cases where one of the inputs is a derivatives of another: in this case the full "path" to the shorter step is given.

The problem arrives if the total "path" lengths for multiple desired test cases do not match, for example, `compiled-with-debug-info` started chain is always one step shorter than the one that goes through `compiled-stripped`. As such, when doing a basic filter on two derived inputs, no test with `debug-info` will ever be run.

The second problem with tags is exactly the same. `native-dynamic` test case is always one tag shorter compared to `static` cases with given architectures, so, similarly, when multiple inputs overlap, the `dynamic` test cases never make it thorough.

P. S. There's also a minor fix pushing all the tags and variables through a `set` when merging them - to avoid duplication.

### DIFF
--- a/revng-test-configure
+++ b/revng-test-configure
@@ -59,8 +59,9 @@ def most_similar_tuples(inputs):
     if len(inputs) == 1:
         return list(product(*inputs))
 
-    result = []
+    fully_compatible = []
 
+    most_similar = []
     maximum_similarity = 0
 
     # Get the most similar set of tuples, i.e., those that share the most
@@ -79,15 +80,26 @@ def most_similar_tuples(inputs):
 
         if similarity > maximum_similarity:
             maximum_similarity = similarity
-            result = []
+            most_similar = []
 
         if similarity == maximum_similarity:
-            result.append(candidate_set)
+            most_similar.append(candidate_set)
+
+        for candidate in candidate_set:
+            if candidate in common_targets:
+                fully_compatible.append(candidate_set)
+                break
+
+    # For now, discard the `most_similar` list in case there are any perfect
+    # matches (perfect match is when all the inputs in the current candidate set
+    # derive from a single input present in the same set)
+    # TODO: consider if we still want to keep it.
+    selected = fully_compatible if len(fully_compatible) != 0 else most_similar
 
     # Further refine based on common tags
     maximum_similarity = 0
-    result2 = []
-    for candidate_set in result:
+    result = []
+    for candidate_set in selected:
         candidate_tags = [set(candidate.tags)
                           for candidate
                           in candidate_set]
@@ -98,12 +110,12 @@ def most_similar_tuples(inputs):
 
         if similarity > maximum_similarity:
             maximum_similarity = similarity
-            result2 = []
+            result = []
 
         if similarity == maximum_similarity:
-            result2.append(candidate_set)
+            result.append(candidate_set)
 
-    return result2
+    return result
 
 def sort_list(entries, key, dependencies):
     # Collect entries by key and id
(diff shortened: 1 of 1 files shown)


Cvelth:
An example (cc: @aleclearmind):

Current behaviour on `develop` when triggering the [`revng.cross-relations` test](https://github.com/revng/revng/blob/develop/share/revng/test/configuration/revng/yields.yml#L50):
```
calc-x86-64-static-revng.cross-relations-$HASH.yml # stripped
calc-i386-static-revng.cross-relations-$HASH.yml # stripped
calc-s390x-static-revng.cross-relations-$HASH.yml # stripped
calc-mips-static-revng.cross-relations-$HASH.yml # stripped
calc-mipsel-static-revng.cross-relations-$HASH.yml # stripped
calc-arm-static-revng.cross-relations-$HASH.yml # stripped
calc-aarch64-static-revng.cross-relations-$HASH.yml # stripped
```

After `test-configure: prefer fully compatible tuples`, it becomes:
```diff
calc-x86-64-static-revng.cross-relations-$HASH.yml # stripped
calc-i386-static-revng.cross-relations-$HASH.yml # stripped
calc-s390x-static-revng.cross-relations-$HASH.yml # stripped
calc-mips-static-revng.cross-relations-$HASH.yml # stripped
calc-mipsel-static-revng.cross-relations-$HASH.yml # stripped
calc-arm-static-revng.cross-relations-$HASH.yml # stripped
calc-aarch64-static-revng.cross-relations-$HASH.yml # stripped
+calc-x86-64-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-i386-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-s390x-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-mips-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-mipsel-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-arm-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-aarch64-static-revng.cross-relations-$HASH.yml # with debug-info
```

And after `test-configure: prefer fully compatible tags`:
```diff
calc-x86-64-static-revng.cross-relations-$HASH.yml # stripped
calc-i386-static-revng.cross-relations-$HASH.yml # stripped
calc-s390x-static-revng.cross-relations-$HASH.yml # stripped
calc-mips-static-revng.cross-relations-$HASH.yml # stripped
calc-mipsel-static-revng.cross-relations-$HASH.yml # stripped
calc-arm-static-revng.cross-relations-$HASH.yml # stripped
calc-aarch64-static-revng.cross-relations-$HASH.yml # stripped
+calc-dynamic-native-revng.cross-relations-$HASH.yml # stripped
calc-x86-64-static-revng.cross-relations-$HASH.yml # with debug-info
calc-i386-static-revng.cross-relations-$HASH.yml # with debug-info
calc-s390x-static-revng.cross-relations-$HASH.yml # with debug-info
calc-mips-static-revng.cross-relations-$HASH.yml # with debug-info
calc-mipsel-static-revng.cross-relations-$HASH.yml # with debug-info
calc-arm-static-revng.cross-relations-$HASH.yml # with debug-info
calc-aarch64-static-revng.cross-relations-$HASH.yml # with debug-info
+calc-dynamic-native-revng.cross-relations-$HASH.yml # with debug-info
```

aleclearmind:
This PR reason to be has been fixed in `feature/fix-test-list`, which has been recently merged.

Closing, but thanks for bringing this up!

PR #17, opened 2023-03-21
Start testing 32-bit microsoft ABIs

Cvelth:


share/revng/test/tests/abi/include/common.h
```diff
@@ -65,18 +65,18 @@ _Static_assert(sizeof(uint8_t) == 1, "A type with size == 1 is required.");
...
-#define PRINT(TYPE, POINTER)                         \
-  do {                                               \
-    printf("      - Type: " #TYPE " # size = 0x%x\n" \
-           "        Bytes: ",                        \
-           sizeof(TYPE));                            \
-    MAKE_PRINT_HELPER(TYPE, POINTER, local_helper);  \
-    PRINT_BYTES(TYPE, local_helper);                 \
+#define PRINT(TYPE, POINTER)                        \
+  do {                                              \
+    printf("      - Type: " #TYPE " # size = %u\n"  \
```
aleclearmind: Rename this commit to "Emit type size in decimal"

scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -50,6 +54,8 @@
       bl {{ function_name }}
     return_from_function: |
       ret
+    return_from_function_with_cleanup: |
+      ret
```
aleclearmind: You're not doing any cleanup here, similar below

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
```diff
@@ -65,7 +65,9 @@ void describe_/*( name )*/(void) {
       /*( fill_stack_with_random_data )*/
       /*( fill_registers_with_random_data )*/
       /*( call_a_function(name) )*/
+#ifndef CALLEE_CLEANED_ARGUMENTS
       /*( restore_stack )*/
```
aleclearmind: A more appropriate name would be `cleanup_stack_arguments`

share/revng/test/configuration/revng-qa/for-detect-abi.yml
```diff
@@ -65,6 +65,57 @@ tags:
...
+        - -DSTRUCT_SIZE_LIMIT=0
+
+  - name: i386-fastcall-microsoft-abi
+    implies: [i386]
+    variables:
+      ABI_NAME: Microsoft_x86_fastcall
+      CFLAGS:
+        - -DABIDEF='__attribute__((ms_abi, fastcall))'
+        - -DSCALAR_SIZE_LIMIT=2
+        - -DSTRUCT_SIZE_LIMIT=0
```
aleclearmind: 1. These are rather obscure and need to be documented somewhere. At least at the top of this file.
2. Not necessarily for now, but we should consider whether we want to move all of this "configuration" into a `.h` file.
   i.e., Just `-DABI_Microsoft_x86_fastcall` and then in a `.h` file have:
   ```c
   #ifdef ABI_Microsoft_x86_fastcall
   #define ABIDEF __attribute__((ms_abi, fastcall))
   #define SCALAR_SIZE_LIMIT 2
   #define STRUCT_SIZE_LIMIT 0
   #elif ...
   ```
   These configurations are not "test configuration" but more like "ABI properties definitions", and I feel like they belong somewhere else.

aleclearmind (changes requested):

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Yeah, it's my bad on choosing the template name. Maybe I should actually modify the template to detect if this value was provided and fall back onto `return_from_function` in the cases where it was not 🤔 

What's your stance?

share/revng/test/configuration/revng-qa/for-detect-abi.yml
Cvelth: In a perfect world, we'd have a script to read our yaml definitions and figure these out from it. But currently that would cause a dependency cycle (since yamls are in revng and we _really_ don't want these ugly generators down there) - so I opted out to temporarily hand-write them until we come up with a better solution.

aleclearmind: We don't want to depend on revng.
I'd go for the header.

Cvelth: I guess header is still an improvement compared to this. Even if not ideal.

scripts/abi_artifact_generator/config/architectures.yml
aleclearmind: 1. I think it's a mistake if we ever use `return_from_function_with_cleanup` for an architecture for which we don't want to support callee-cleaned up ABIs. Can we put an invalid instruction to prevent assembling it entirely?
2. Why are we hardcoding the amount of stuff to pop from the stack? `ret $8`

Cvelth: 1. Well yeah, this is only ever used for ABI for which calle-cleared macro is explicitly set. I'd say, an invalid instruction does indeed make a lot of sense.
2. It's a pointer, which is a constant for a given architecture. If you prefer, I can give it a name, but it's an ugly hack either way - named or not.

aleclearmind: > It's a pointer, which is a constant for a given architecture. If you prefer, I can give it a name, but it's an ugly hack either way - named or not.

This means we can only test having a pointer-sized stack argument (for callee-saved ABIs)? Can't we leave a placeholder and fill it in later on?

Cvelth: This is specifically for return value stuff. This instruction is only ever inserted when a function returns stuff by taking a pointer to the stuff where it has to be placed as a stack argument.

The common case with argument, luckily, doesn't require us to do anything special since `setjmp`/`longjmp` pair handles cleanup for us.

aleclearmind: Ah so this is for the wrapper function that pushed the argument?

Please add a comment at the top of the file what these snippets are for and document each field.

Cvelth: More like "wrapped" function that readies the return value, but yeah, popping one register is all it needs.

Documentation - sounds good.

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
Cvelth: I went with `CALLEE_CLEANS_STACK_ARGUMENTS_UP` macro instead. If that sounds good to you, feel free to close this thread.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: The files started to look ugly, so I moved the documentation outside - into a dedicated markdown file. And while at it, I also added the rest of the config in as well - for completeness.

See `scripts/abi_artifact_generator/config/readme.md`.

share/revng/test/tests/abi/include/common.h
Cvelth: Done.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Here's the link straight to it: https://github.com/revng/revng-qa/tree/f0d5b2c21457434a1c441aa1f3b276782399d25d/scripts/abi_artifact_generator/config

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
aleclearmind: I was talking about `restore_stack`.

Cvelth: Ah, my bad! I was under the assumption that it is used elsewhere, but no, it looks like I removed the other usage at some point because it wasn't necessary.

scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -73,6 +79,8 @@
       bl {{ function_name }}
     return_from_function: |
       bx lr
+    return_from_function_with_cleanup: |
+      .word 0xe7f0def0
```
aleclearmind: What's this? A comment at least!

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
aleclearmind: What do you mean?
I was suggesting to rename `cleanup_stack_arguments` into  `restore_stack`?

aleclearmind (changes requested):

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: I just googled around for arm's invalid instructions. And this is the most robust thing I found.

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
Cvelth: Yeah, sorry. I meant that `restore_stack` used to have more usage in the past, which is why I was reluctant to consider that. But it looks like all but this one are gone already - so I'm fine with the new name.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Added the comment.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
Cvelth: The header is ready, please take a look: `share/revng/test/tests/abi/abi.h`.

scripts/abi_artifact_generator/templates/describe_functions.inc.jinja_template
Cvelth: Renamed it.

scripts/abi_artifact_generator/config/architectures.yml
aleclearmind: Do not link commits, I told you many times.
I'll continue from `readme.md`.

scripts/abi_artifact_generator/config/readme.md
```diff
@@ -0,0 +1,99 @@
+# Documentation for the configuration options provided by this generator
```
aleclearmind: Please start off stating the goal of the tool and how to invoke it

share/revng/test/tests/abi/abi.h
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+#define UNLIMITED 65536
+
+// This file contains a dirty trick. Ideally we'd want to automagically import
+// these values from the corresponding abi `yaml` files contained in revng, but
+// for now manually copying the values here will do.
+
+#if defined(ABI_SystemV_x86_64)
+#define ABIDEF
```
aleclearmind: What is this for?

share/revng/test/tests/abi/abi.h
```diff
@@ -0,0 +1,93 @@
...
+#define MaximumGPRsPerAggregateArgument UNLIMITED
+#define MaximumGPRsPerScalarArgument UNLIMITED
+#define MaximumGPRsPerAggregateReturnValue 2
+#define MaximumGPRsPerScalarReturnValue 2
+#elif defined(ABI_SystemV_x86)
```
aleclearmind, line 31: Add a blank line before each ABI description

aleclearmind (changes requested):

share/revng/test/tests/abi/abi.h
Cvelth: Yeah, empty definitions are actually useless. I'll backport the updated `abi.h` here after you review it on the other PR.

scripts/abi_artifact_generator/config/readme.md
aleclearmind, line 1: Rename to `README.md`

scripts/abi_artifact_generator/config/readme.md
Cvelth: Something like this?
```markdown
The goal of this generator is to have a combined interface for generating test for each of the testeed architectures by taking advantage of multiple templates with varying levels of genericness.

This file explains the said configuration in multiple stages:
* [the values shared between all the tests](#commonyml)
* [the way to specify the list of functions for the tests to be based on](#functionsyml)
* [the way to add/modify a specific architecture to the testing suite](#architecturesyml)
```

share/revng/test/tests/abi/abi.h
Cvelth: The extra ones are gone.

share/revng/test/tests/abi/abi.h
Cvelth: Done.

aleclearmind (changes requested):

scripts/abi_artifact_generator/config/readme.md
Cvelth: Done.

scripts/abi_artifact_generator/config/readme.md
aleclearmind: Uhm, it can be better, we need some more in depth documentation on what we're trying to achieve.

We need to explain with examples what we do and why. One example is worth a thousand words.
We need to explain that we initialize the state with random data and then verify that the final state contains the data we expected to be present in certain locations.

Explain the thing end-to-end with examples.

Cvelth: On it.

revng-test-configure
```diff
@@ -97,10 +97,10 @@ def most_similar_tuples(inputs, command_type):
...
     log(json.dumps([[x.path for x in input] for input in result], indent=2))
 
-    for input_index, input in enumerate(inputs):
-        for target_index, target in enumerate(input):
-            if (input_index, target_index) not in used_inputs:
-                error(f"Target {target.path} has not been used!")
+    # for input_index, input in enumerate(inputs):
+    #     for target_index, target in enumerate(input):
+    #         if (input_index, target_index) not in used_inputs:
+    #             error(f"Target {target.path} has not been used!")
```
aleclearmind: How can I help?
Can you describe the issue?

scripts/abi_artifact_generator/templates/functions.h.jinja_template
```diff
@@ -48,6 +48,22 @@ typedef struct {
 
 /*{ endfor -}*/
 
+// There's a slight difference in the way this pragma impacts the structs
+// between msvc and gcc, but it shouldn't be relevant in our case: since all
+// we care about is that these structs are emitted as "unnaturally aligned",
+// without any regard to their specific size.
```
aleclearmind, line 66: This doc is missing the "why don't we care?" part. How can we predict what will happen if we don't know the size of the struct?
Maybe it affects only the size and not the starting offset?

scripts/abi_artifact_generator/templates/functions.h.jinja_template
```diff
@@ -48,6 +48,22 @@ typedef struct {
...
+#pragma pack(push, 1)
```
aleclearmind, line 75: I'd prefer `__attribute__((packed))`, which we usually use.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
```diff
@@ -77,6 +82,11 @@ sources:
       - [i386-regparm-3-systemv-abi]
       - [i386-regparm-2-systemv-abi]
       - [i386-regparm-1-systemv-abi]
+      - [x86-64-systemv-abi, intel-asm-syntax]
+      - [i386-systemv-abi, intel-asm-syntax]
+      - [i386-regparm-3-systemv-abi, intel-asm-syntax]
+      - [i386-regparm-2-systemv-abi, intel-asm-syntax]
+      - [i386-regparm-1-systemv-abi, intel-asm-syntax]
```
aleclearmind: 1. AFAIU this is to let the compiler know that inline assembly will use Intel syntax, correct?
2. Can't we simply put `.intel_syntax;` in the inline assembly? It would save us these non-local changes.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
```diff
@@ -0,0 +1,53 @@
...
+/*{- if current_style == 'gcc' -}*/
+.global setup_/*( name )*/
+setup_/*( name )*/:
+/*{- elif current_style == 'msvc' -}*/
+setup_/*( name )*/ proc
+/*{- endif }*/
+  /*( save_return_address )*/
+  /*( fill_stack_with_random_data )*/
+  /*( fill_registers_with_random_data )*/
+  /*( call_a_function("test_" + name) )*/
```
aleclearmind: As far as I can see, `save_return_address`, `fill_stack_with_random_data`, `fill_registers_with_random_data` are used only once, here.

Can we merge them and call it like "setup_prologue"?

We're missing docs, but `save_return_address` does not convey very well that it will be called at the entry of the function.

More in general, can you look into how much can we squash together things that always go together? Like `restore_return_address` and `return_from_function`.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
```diff
@@ -0,0 +1,53 @@
+/*#-
```
aleclearmind, line 1: Can you pass me the set of generated files?
Looking at the generated file and tracking it back to the template is much more effective to get gist of what we're doing.

NOTETOSELF: review the templates.

scripts/abi_artifact_generator/templates/test_list.txt.jinja_template
```diff
@@ -0,0 +1,4 @@
+/*{- for name, arguments in argument_functions.items() -}*//*( name )*/ /*{ endfor -}*/
+/*{- for name, arguments in return_value_functions.items() -}*//*( name )*/ /*{ endfor }*/
+randomized_state printable_location
+/*( lfsr_seed )*/ /*( iteration_count )*/ /*( stack_byte_count )*/ /*( generated_byte_count )*/
```
aleclearmind: What is this for?

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+        assert len(extra_values) == 4
+        (
+            result.seed,
+            result.iteration_count,
+            result.stack_byte_count,
+            result.generated_byte_count,
+        ) = extra_values
+
+    with open(str(arguments.input_file), "r+") as input_file:
+        with closing(mmap(input_file.fileno(), 0, access=ACCESS_READ)) as input_data:
```
aleclearmind: Just read the file, why are you using `mmap`? Early optimization?

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+        data,
+    )
+    if match:
+        return (
+            match[1].decode("utf-8"),  # size
+            match[2].decode("utf-8"),  # vma
+            match[4].decode("utf-8"),  # offset
+        )
+    else:
+        sys.exit(2)
```
aleclearmind: Print an error message

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+def accept_arguments():
+    parser = argparse.ArgumentParser(
+        description="A quick disassembly analyzer that emits some addresses as a C header.",
+    )
+
+    parser.add_argument(
+        "input_file",
+        type=str,
+        help="the file to read disassembly from",
+    )
```
aleclearmind: Use `sys.stdin`

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+    parser.add_argument(
+        "list_of_symbols_to_extract",
+        type=str,
+        help="the file listing the interesting symbols",
+    )
+    parser.add_argument(
+        "output_file",
+        type=str,
+        help="the file to write the results to",
+    )
```
aleclearmind: Use `stdout`

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+import re
+import sys
+from contextlib import closing
+from dataclasses import dataclass
+from mmap import ACCESS_READ, mmap
+
+
+def accept_arguments():
+    parser = argparse.ArgumentParser(
+        description="A quick disassembly analyzer that emits some addresses as a C header.",
```
aleclearmind: 1. What is the input of this script? binutils' `objdump`? `llvm-objdump`?
2. Does this work with PE/COFF too?

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+
+class Function(MemoryLocation):
+    def __init__(self, name: str, address: str, entry_point: str):
+        self.name = name
+        self.address = address
+        self.entry_point = entry_point
+        self.callsites: list[CallSite] = []
+
+
+class Gathered:
```
aleclearmind, line 110: Gathered?

share/revng/test/tests/abi/runner.c
```diff
@@ -0,0 +1,226 @@
...
+
+int main(int argc, char **argv) {
+  /* Setup signal handling */
+  struct sigaction action = { .sa_sigaction = &handler,
+                              .sa_flags = SA_SIGINFO };
+  sigemptyset(&action.sa_mask);
+  sigaction(SIGILL, &action, NULL);
+
+  if (argc != 2)
+    return 1;
```
aleclearmind, line 190: Emit meaningful error message upon error

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -0,0 +1,398 @@
...
+        + b"([\*\.\w]+)\s+"  # section symbol belongs to
+        + b"([a-fA-F0-9]+)\s+"  # either alignment or size of the symbol
+        + name_bytes  # name of the symbol
+        + b"\n",
+        data,
+    )
+    if match:
+        return match[1].decode("utf-8")  # address
+    else:
+        sys.exit(1)
```
aleclearmind: Emit meaningful error messages

share/revng/test/tests/abi/runner.c
```diff
@@ -0,0 +1,226 @@
...
+
+  /* load the "inner" binary */
+  int binary_file = open(argv[1], O_RDONLY);
+  struct stat binary_file_stats;
+  fstat(binary_file, &binary_file_stats);
```
aleclearmind: Assert return codes are those you expect of all POSIX calls.

share/revng/test/tests/abi/runner.c
```diff
@@ -0,0 +1,226 @@
...
+  saved.before_the_call = ACCESS_A_WORD(function->callsites[0].before);
+  ACCESS_A_WORD(function->callsites[0].before) = ILLEGAL_INSTRUCTION_WORD;
+  saved.after_the_return = ACCESS_A_WORD(function->callsites[0].after);
+  ACCESS_A_WORD(function->callsites[0].after) = ILLEGAL_INSTRUCTION_WORD;
+
+  void (*entry_point)(void) = (void (*)(void))(size_t) function->entry_point;
+  if (setjmp(buffer) == 0) {
+    entry_point();
+    longjmp(buffer, 1);
+  }
```
aleclearmind, line 103: Are we evil? `:D`

share/revng/test/configuration/revng-qa/compile.yml
```diff
@@ -1,11 +1,12 @@
 tags:
   - name: stripped
   - name: with-debug-info
+  - name: has-special-compilation-requirements
```
aleclearmind: This really looks like a hack. Can you detail so we can try to get around it?

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
```diff
@@ -0,0 +1,83 @@
+tags:
```
aleclearmind: NOTETOSELF: review this commit

aleclearmind (changes requested):

revng-test-configure
Cvelth: This is the context behind this comment: https://mattermost.rev.ng/revng/pl/khqff83zbjd5xcdh57rdqpcu6e

I haven't looked into it yet, but I can if you'd rather I do it.

Cvelth: (only the first couple of messages in that thread tree, we go on a tangent after that)

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: We don't care because we get the size as `sizeof(struct)` - so whatever compiler decides to do with it - it's consistent, so size everywhere is going to be as well. And as long as it also considers said struct unnaturally aligned - we're happy.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: MSVC doesn't support it. Only pragmas there. I'd rather keep consistent approach, even if it's slightly harder to work with.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
Cvelth: 1. More like to make the compiler use exclusively `intel syntax`, I don't remember the specifics, but I have read somewhere that it's a bad idea to mix them - there are some minor incompatibilities.

2. First of all, we don't use inline assembly anymore. But also, I kind of prefer how explicit flag feels.

One option would be to try and move this flag higher up or something instead of introducing a separate tag for it or something.

P. S. I don't have any strong opinions on this. As long as the compiler is happy - so am I.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Thanks for noticing! It's a bit hard for me to keep track of all of this since it changed so much lately, but yes, the less templates we have, the easier they are to work with.

scripts/abi_artifact_generator/templates/test_list.txt.jinja_template
Cvelth: This is the trick I used to avoid having to templatize `runner.c`. It passes all the extra data it needs as part of the header it takes as the main input.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: This way the file stays as binary, so it's easier to run binary regexes on it. Either way, I didn't  put much thought into this, this is inspired by the very thing that came up from stack overflow about `regex`ing big files.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Not much thought into the name. Since the script is `gathering` stuff, this is the stuff it has `gathered` as the result. I guess something like `output` would be a better name.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: I want to reuse our usual architecture tags (`i386`, `arm`, you get the point), but as soon as something is tagged with them - it gets compiled normally. So it was either a separate set of abi-specific tags (🤢) or something like this to prevent select sources from being sweeped.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
aleclearmind: Yes, that's what the comment says to. But why we're happy? What happens with unaligned `struct`s downstream? Don't we want to test `struct` offsets?

scripts/abi_artifact_generator/templates/functions.h.jinja_template
aleclearmind: Makes sense.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
aleclearmind: This change is not present in the final diff. Why is that?
Please squash fixup'd commits.

share/revng/test/tests/abi/gather_symbols.py
aleclearmind: This is not big nor binary, let's drop `mmap` to avoid confusion.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: > But why we're happy?

We never test internals of the structs. When forcing them to be unaligned, compiler is definitely putting them on the stack (he doesn't want to deal with it either). So, in essence, these tests are only there so we can verify that function types layouts are aware that some stuff is bound to only use stack.

As for verification, we can just compare them bytewise. Unlike return values, here it doesn't seem to cause any extra problems.

> Don't we want to test `struct` offsets?

what for? As long as struct is where we expect it to be, aren't we happy? I can't imagine compiler doing something too weird with the internals just because it's an argument or something. 

In short, if the bytes we want to see are where we are looking for them - we are happy, no matter compiler thinks about their type or whatnot.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
Cvelth: Looks like I lost it somewhere along the way, now I'm really confused about why it even works. There's probably a rogue flag hardcoded somewhere or something. I'll investigate tomorrow.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: 1. I'd argue ~1.4Mb (arm one) of text can be considered quite big.

2. Sure, but I'd still rather keep it as a binary file, this makes regexes more efficient AND easier to work with, since specifics of python strings don't come into play. Which, if I recall correctly, is relevant with multiline matches.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
aleclearmind: OK, please add to the comment the following part of your answer:

> When forcing them to be unaligned, compiler is definitely putting them on the stack (he doesn't want to deal with it either). So, in essence, these tests are only there so we can verify that function types layouts are aware that some stuff is bound to only use stack

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: No need, I'll build locally. It's simpler.

Cvelth: Should I maybe make a graph or something of how all the pieces fit together?

On top of making your review easier, it should also serve a good way to bring all the docs together by highlighting how different pieces interact.

share/revng/test/tests/abi/gather_symbols.py
aleclearmind: You're really early optimizing here. There's a `re.MULTILINE` flag for regexps you can pass, no need to use bytes, which is kinda confusing for a text file. If you really want to use bytes, open it as binary, no need for `mmap`. Let's drop `mmap`.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
```diff
@@ -0,0 +1,53 @@
...
+This template file is distributed under the MIT License. See LICENSE.md for details.
```
aleclearmind, line 2: Make sure each `$MY_NAME.jinja_template` file generates a `$MY_NAME` file.
Currently this file generates `{msvc,gcc}_setup_arguments.S`.
Use directories to further organize the files.

scripts/abi_artifact_generator/templates/types.h.jinja_template
```diff
@@ -0,0 +1,42 @@
...
+#ifdef __SIZEOF_INT128__
+#define INT128_IS_SCALAR 1
+#else
+#define INT128_IS_SCALAR 0
+#endif
+
+#define IS_SCALAR(TYPE, VALUE) \
+  _Generic((VALUE), \
+    /*{- for type in supported_primitives }*/
+           /*( type )*/: 1, \
```
aleclearmind, line 30: Why `1` instead of `true`?

scripts/abi_artifact_generator/templates/types.h.jinja_template
```diff
@@ -0,0 +1,42 @@
...
+
+/*(- generation_notice )*/
+
+#ifdef __SIZEOF_INT128__
+#define INT128_IS_SCALAR 1
+#else
+#define INT128_IS_SCALAR 0
+#endif
+
+#define IS_SCALAR(TYPE, VALUE) \
```
aleclearmind, line 27: What's definition of *scalar* here?
Some `struct`s seem to be classified as scalars.
Add a comment.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: Also, add a comment in the output so it's clear what's going on.

Something like

```patch
diff --git b/./abi_artifact_generated/x86_64/gcc_setup_arguments-original.S a/./abi_artifact_generated/x86_64/gcc_setup_arguments.S
index cce88c0..2cbeb67 100644
--- b/./abi_artifact_generated/x86_64/gcc_setup_arguments-original.S
+++ a/./abi_artifact_generated/x86_64/gcc_setup_arguments.S
@@ -10,8 +10,11 @@
 
 .global setup_single_8_bit_argument
 setup_single_8_bit_argument:
+  ; Save return address
   pop rax
   mov qword ptr [saved_return_address], rax
+
+  ; Populate the stack with random data
   mov rax, qword ptr [randomized_state + 616]
   push rax
   mov rax, qword ptr [randomized_state + 608]
```

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: The idea was `${STYLE_NAME_IF_ANY}_${MY_NAME}`, but I guess directories work too.

scripts/abi_artifact_generator/templates/types.h.jinja_template
Cvelth: Because that's `c` and `true` involves including `stdbool.h` (or whatever it's called), which caused problems with msvc.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
```diff
@@ -0,0 +1,53 @@
...
+extern saved_return_address: /*( intel_register_type )*/
+
+/*{ for name, arguments in argument_functions.items() -}*/
+extern test_/*( name )*/: proc
+/*{ endfor }*/
+
+.code
+/*{- endif }*/
+
+/*{ for name, arguments in argument_functions.items() }*/
```
aleclearmind, line 32: The body of this function is always the same except for the address of the callee, correct?

Add a comment stating that we need to replicate the body of this function for each callee so that we can fully control the CPU state and we don't have to sacrifice a register to perform the indirect call.

It looks to me that there's still some opportunity to outline some code and emit it just once, e.g., the part that populates the stack.

```
setup_single_8_bit_argument:
  pop rax
  mov qword ptr [saved_return_address], rax
  call populate_stack
  mov r15, qword ptr [randomized_state + 104]
  mov r14, qword ptr [randomized_state + 96]
  mov r13, qword ptr [randomized_state + 88]
  mov r12, qword ptr [randomized_state + 80]
  mov r11, qword ptr [randomized_state + 72]
  mov r10, qword ptr [randomized_state + 64]
  mov r9, qword ptr [randomized_state + 56]
  mov r8, qword ptr [randomized_state + 48]
  mov rdi, qword ptr [randomized_state + 40]
  mov rsi, qword ptr [randomized_state + 32]
  mov rdx, qword ptr [randomized_state + 24]
  mov rcx, qword ptr [randomized_state + 16]
  mov rbx, qword ptr [randomized_state + 8]
  mov rax, qword ptr [randomized_state + 0]
  call test_single_8_bit_argument
  ; ...
```

Maybe we could go even a bit further.
Anyway, if you think this is viable, **do not do this now** but add a comment.

scripts/abi_artifact_generator/templates/types.h.jinja_template
Cvelth: > Some `struct`s seem to be classified as scalars

what do you mean? Where? Do you, maybe, mean the pointers?

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
```diff
@@ -0,0 +1,53 @@
...
+
+/*{ if current_style == 'gcc' -}*/
+#include "../abi.h"
+
+#if defined(TARGET_x86_64) || defined(TARGET_i386)
+.intel_syntax noprefix
+#endif
+/*{- elif current_style == 'msvc' -}*/
+ifndef rax
+.model flat, c
```
aleclearmind, line 18: What's this? Comment.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,144 @@
+/*#-
```
aleclearmind, line 1: 1. Add a comment at the top of each template about what is the goal of the file.
2. AFAIU this decodes a global buffer containing data dumped by the various functions. `printers.c` does not convey this very well. Can we give a proper name to the buffer above and call all of this `decode_$NAME.c`?

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,144 @@
...
+  do {                                                                         \
+    TYPE NAME; (void) NAME;                                                    \
+    if (sizeof(TYPE) <= ARGUMENT_SIZE_LIMIT(TYPE, NAME) && !IS_PACKED(TYPE)) { \
+      PRINT(TYPE, (LOCATION));                                                  \
+      current_offset += sizeof(TYPE);                                          \
+    } else {                                                                   \
+      POINTER(TYPE, (LOCATION));                                                \
+      current_offset += sizeof(TYPE *);                                        \
+    }                                                                          \
+  } while (0)
```
aleclearmind: Hoist in one of the headers above or in a new one?
Same with functions at the bottom.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,144 @@
...
+
+#define ARGUMENT(TYPE, NAME, LOCATION) \
+  do {                                                                         \
+    TYPE NAME; (void) NAME;                                                    \
+    if (sizeof(TYPE) <= ARGUMENT_SIZE_LIMIT(TYPE, NAME) && !IS_PACKED(TYPE)) { \
+      PRINT(TYPE, (LOCATION));                                                  \
+      current_offset += sizeof(TYPE);                                          \
+    } else {                                                                   \
+      POINTER(TYPE, (LOCATION));                                                \
+      current_offset += sizeof(TYPE *);                                        \
```
aleclearmind: Macros that assume the existence of some variable in the caller are evil!
Add a macro argument, we might even consider passing `&current_offset` in to be extra explicit about the fact that it will be manipulated within the macro.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,144 @@
...
+  puts("    ReturnValue:");
+  PRINT(type, (output + 8));
+  puts("    ExpectedReturnValue:");
+  PRINT(type, input);
+}
+
+/*{ endfor -}*/
+
+/* Dispatches to the current printer call based on the test name */
+void print(const char *name, const uint8_t *input, const uint8_t *output) {
```
aleclearmind, line 112: Supposing we dump on disk `output`, is there something preventing us from doing all of this from Python?
We seem to have all the meta information required to do this in Python.

If so, please add a comment that this might be a viable future direction.

scripts/abi_artifact_generator/templates/setup_return_values.inc.jinja_template
```diff
@@ -0,0 +1,35 @@
...
+  typedef /*( return_value_type )*/ return_value_type;
+
+  // Call the tested function.
+  return_value_type returned = test_/*( name )*/();
+
+  // Write the pointer to the received values into a safe location,
+  *(return_value_type **) printable_location = &returned;
+
+  // Then write the value to it as well.
+  *(return_value_type *) (printable_location + 8) = returned;
```
aleclearmind, line 31: Are all the types we deal with (`struct`s in paritcular) always "copy assignable"? I expected we had to use `memcpy` here.

share/revng/test/configuration/revng-qa/compile.yml
aleclearmind: So you want to inherit the properties of the, say, `arm` tag (e.g., `TRIPLE: armv7a-hardfloat-linux-uclibceabi-`) but not have the source be fed to `revng-qa.compiled*` commands. Correct?

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
```diff
@@ -0,0 +1,83 @@
...
+
+  - type: revng-qa.abi-test-extracted-addresses
+    from:
+      - type: revng-qa.disassembled-copy
+      - type: source
+        filter: abi-test-list
+      - type: source
+        filter: abi-test-symbol-gathering-script
+    suffix: .h
+    command: python3 ${INPUT3} ${INPUT1} ${INPUT2} ${OUTPUT}
```
aleclearmind: This command is overly complicated. I'd like to reduced it to have a single argument: `revng-qa.disassembled-copy`.
1. About `source` filtered by `abi-test-list`: the contents of these files seems to be entirely deterministic:
   1. The list of functions is fixed for all architectures and in any case we can recover them by the `objdump`.
   2. The `lfsr_seed` can be `0xDEADBEEF`.
   3. `iteration_count` is fixed.
   4. `stack_byte_count` is fixed.
   5. `generated_byte_count` can be bumped to a big enough value and become fixed. Say 2048.
2. About `source` filtered by `abi-test-symbol-gathering-script`: just hardcode the script name. I'm really confused why you decided to do this.
3. Do not create `commands` whose output is used only once. `revng-test-configure` is not a build system, we're not trying to save time on incremental builds. This is not the first time you try to use it this way, please refrain from doing this, it brings extra complexity and no benefits.
    If a command's output is used only in a single other command, squash it into it. For instance `revng-qa.disassembled-copy` can be squashed into `revng-qa.abi-test-extracted-addresses`. Review all other new commands that have been introduced as well.

share/revng/test/tests/abi/runner.c
```diff
@@ -0,0 +1,226 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "gathered_symbols.h"
+#include "ucontext_wrapper.h"
+
```
aleclearmind, line 44: Add a comment briefly explaining the idea of this program. Like: we have some metadata from a `.h` file, we `mmap` a file, we used the metadata to introduce breakpoints here and there, dump the state in the signal handler and restore the original instruction, resume execution, etc.

Putting this documentation somewhere else just increases the risk of it gettting outdated.

aleclearmind (changes requested):

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: No, let's focus on merging.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: Yeah, but with so many templates, predictability makes life easier.

scripts/abi_artifact_generator/templates/types.h.jinja_template
aleclearmind: OK, add a comment for future generations.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Yep, but the big problem is that we cannot really afford to use a register for this, so it needs to either be some weird trick, or self-modifying code.

Either way, comment is a good idea.

scripts/abi_artifact_generator/templates/types.h.jinja_template
aleclearmind: Right, my bad!

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: It declares calling convention, `flat, c` means `cdecl`.

Which is one of the big problem we'd have to solve if we plan to go full PE/COFF binary. This was easily avoidable in the Frankenstein approach.

Either way, out of scope for the current PR. Will do the comment though.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: Makes sense. Thanks for noticing.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: I'd guess qemu is my main worry. By using C we are getting a fully self-contained binary. So it's as simple as running it to test all the function. Making it more script heavy == quite a few more explicit `qemu` invocations, which will add up really fast.

scripts/abi_artifact_generator/templates/setup_return_values.inc.jinja_template
Cvelth: I prefer not to rely onto `libc` if we can avoid it, which we currently do.

Also, this way it's more natural - since we leave it to the compiler to copy the object "naturally", which in my experience sometimes leaves out the padding (which is a problem, but I'd say not a big one).

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: Yes, exactly.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: > we cannot really afford to use a register for this

What do you mean? What would be wrong with my example above?

If we just add a call + ret, in x86-64 means we use (and free) a stack slot, in other architecture we clobber the link register, which we were going to clobber anyway a few lines below.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: > one of the big problem we'd have to solve if we plan to go full PE/COFF binary

Please explain

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
Cvelth: 1.1. I have some FUD about that. What if a future version of gcc starts putting in something named `test_X` or whatnot? I'd rather not assume anything about the binary except for the explicit list of tests we purposefully put there.
1.2-4. Works.
1.5. I'm not happy about it, but can live with it.

2. Mostly to avoid hardcoding paths, same as `test_list`. IMO the approach should be the same for both. If you are fine with hardcoding paths here, there's no real reason to drop the list template, IMO.

3. My main reason for doing this is because this way I have an easy access to all the steps I want to see at any point. A test failed? Just pop into the directory it emits stuff into and see everything you need to see without needed to rerun it with a custom "TMP" directory or whatnot.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Closing this then.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
aleclearmind: I'm not following. AFAIU the current workflow is:

1. `runner.c` (which is linked against `printers.c`) produces the "printable" data.
2. `runner.c` invokes functions in `printers.c` to dump in a nice YAML format the printable data. `printers.c` knows how to decode the data thanks to some header generated in Python using some metadata that's available to Python.
3. Someone downstream consumes the YAML.

What prevents us from having `runner.c` dump the raw data and have another script (that uses the same metadata we were using while generating the header) to turn it into a YAML file?

I see not extra invocations of QEMU.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Ah wait, right, I was thinking about it from the ABI perspective. I guess it can work. Sure, it's heavily architecture-specific, but we're already doing architecture-specific stuff here.

scripts/abi_artifact_generator/templates/setup_return_values.inc.jinja_template
aleclearmind: I'm not pushing for `memcpy`, I'm asking if the C standards allows to copy `struct`s. But apparently yes, so, we're happy!

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: For some godforsaken reason, microsoft-flavoured assembly encodes ABI data in. So, a `cdecl` function (`_name`) is not compatible with a `stdcall` function (`_name@4`) which, in turn, is not compatible with a `fastcall` function (`@name@4`).

This "model" is what encodes it for the current file.

In the "Frankenstein" approach, we don't care much, since this doesn't impact instructions - and we only "copy and paste" those into gcc context. If we want a full PE/COFF binary, we'd have to extend templates to actually support all of these fully. Should be doable, but is probably a bit of a PITA.

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
aleclearmind: 1.1 Add a prefix that GCC will never use. But this is really FUD, and we're paying complexity for it. I prefer to fail in the future.
2. What's the different in hardcoding a path in a `sources:` entry or in the script? We already hardcode paths elsewhere.
   Multi input commands need to have a real reason to be used, they trigger all the auto-matching magic and they are a liability. Here we should really have a single input: the binary.
3. We have an issue with easily reproducing test by hand we need to solve, but this is not a solution. This approach leads to a lot of new commands, tags and complexity which we absolutely we need to contain. Squash everything you can. I really really don't want to split a command unless it has multiple outputs (or it takes a very long time).

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: Ah, you meant to cut it at that point. I haven't thought about it. I feel kind of weird about it.

I had a lot of doubts about even separating `printers.c` into a different binary from `functions.c`, but it should be fine as long as they are compiled by the same compiler.

WAIT, now that I'm putting it like that, I'm seeing how it will probably cause problems when MSVC is back into the mix again. We might have no choice but to go with your approach and to have to generate 4 separate binaries (😢).

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
Cvelth: > What's the different in hardcoding a path in a `sources:` entry or in the script?

Having a single spot where anything can be hardcoded (i.e. only the sources) decreases the chance of stuff going out of sync by a lot. And those out-of-sync problem sound like quite a pain to debug if we are ever to run into them. But I guess I'm overthinking it again.

> they trigger all the auto-matching magic and they are a liability

I assume that matching `X options` against `1 option` against `1 option` is practically free no matter what magic it relies on for `AxBxC` stuff. But if you'd rather hardcode paths to avoid that, that works too.

Cvelth: 3. This makes (future) me (who's going to debug these tests going wrong) really sad, but OK.

scripts/abi_artifact_generator/templates/setup_return_values.inc.jinja_template
Cvelth: This! This exact thing makes me so happy I'm a c++ programmer. It never even crossed my mind that there could even exist a world where you giving compiler exact type information is not enough for it to figure out a way to copy an object the right way (tm).

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
aleclearmind: Gotcha.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
aleclearmind: We'll see, I just wanted to caress the idea. Let's get this in!

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
aleclearmind: We already hardcode paths to scripts elsewhere, it's not a problem.

3. I understand, but I think the proper solution would be to make it easier to reproduce a certain command. Right now I run `ninja -vn $MY_FAILING_TEST` and copy paste stuff, but: 1) there are no newlines (hard to read/copy), 2) the `temp -d` stuff is not available in my bash shell, so I have to do tricks. I've some idea about logging `bash -x` output and leaving it somewhere as a reproducer, but we'll see. In any case, I don't want things that could be very simple become very complex for debugging limitations that we can solve some other way.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Done.

scripts/abi_artifact_generator/templates/types.h.jinja_template
Cvelth: Done.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Done.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: Done.

share/revng/test/tests/abi/runner.c
Cvelth: What do you think?

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: Would you be very opposed to postponing this to post-msvc?

If we go with the assembly-less approach for argument too, a lot of this stuff can just go away.

By "postponing" I mean to open a reminder thread in [#20](https://github.com/revng/revng-qa/pull/20) while closing this one.

scripts/abi_artifact_generator/end_to_end.md
```diff
@@ -0,0 +1,308 @@
+# The end to end example of what happens within the testing suite
+
+The suite starts with the generator (see [here](./README.md)), which, in its essense, is just a python scripts that consumes templates (see [here](./templates/)) and configuration files (see [here](./config/)) into the sources.
```
aleclearmind: `s/essense/essence/`, elsewhere too.

scripts/abi_artifact_generator/end_to_end.md
```diff
@@ -0,0 +1,308 @@
+# The end to end example of what happens within the testing suite
```
aleclearmind, line 1: Give a brief overview stating that `randomized_state` is used to initialize the CPU state in a known way and parts of it will be copied into `printable_location` by the code generated by the compiler and then another tool downstream will check if `printable_location` contains what we expect.

scripts/abi_artifact_generator/end_to_end.md
```diff
@@ -0,0 +1,308 @@
...
+  mov rbx, qword ptr [randomized_state + 8]
+  mov rax, qword ptr [randomized_state + 0]
+```
+
+Then it contains a call instruction to the test body itself.
+```asm
+  call test_args
+```
+
+In some cases `longjmp` terminates unless the stack is restored beforehand, so do that:
```
aleclearmind: "terminates" as in "crashes"?

scripts/abi_artifact_generator/templates/README.md
```diff
@@ -0,0 +1,167 @@
...
+The list of data passed from the generator onto the `gather_symbols.py` (see [the relevant docs](../../../share/revng/test/tests/abi/gather_symbols.md)).
+The structure of the file is as follows:
+* List of all the argument test functio names separated by a space in the first line
+* List of all the return value test function names separated by a space in the first line
+* List of all the memory locations `runner.c` accesses in the `mmap'ed binary`.
+* A couple of extra values passed from the generator to the script to keep all the configs in [a single place](../config/common.yml) instead of having to update them all over the place when such an update is needed.
+
+## `types.h`
+
+Due to not wanting to lock ourselves into a single compiler ecosystem, we are not able to rely on the compiler-specific data type logic, which leads to have to generate a "universal" solution based on the data we already have within the generator script. It takes a shape of `_Generic` macro (c11) which provides an unambiguous way to decide whether a specific type will be treated as a scalar or as an aggregate.
```
aleclearmind: `C11` (capital)

aleclearmind (changes requested):

scripts/abi_artifact_generator/end_to_end.md
Cvelth: Yeah, as far as I remember, it was a signal (probably `segfault`, but don't quote me on that).

aleclearmind: What I meant is that the term "terminates" does not imply failure. Use "crashes".

Cvelth: Ah, right. My bad.

scripts/abi_artifact_generator/config/readme.md
Cvelth: Closing this, since there are new related threads.

revng-test-configure
Cvelth: Ping

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Done.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: Ping.

scripts/abi_artifact_generator/templates/setup_arguments.S.jinja_template
Cvelth: Closing this for now.

scripts/abi_artifact_generator/end_to_end.md
Cvelth: Done.

scripts/abi_artifact_generator/end_to_end.md
Cvelth: Done.

scripts/abi_artifact_generator/templates/README.md
Cvelth: Done.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
aleclearmind: Works.

share/revng/test/tests/abi/runner.c
aleclearmind: Looks great!

revng-test-configure
aleclearmind: I cannot reproduce this on the latest commit. Uncommenting works for me. Am I doing something wrong?

Cvelth: I guess the latest version no longer uses whatever combination of `commands` was causing this. Which is good news I'd say.

I'm closing this (also, dropping the commit) and will let you know if I even run into something similar in the future PRs.

share/revng/test/configuration/revng-qa/compile.yml
aleclearmind: I thought about it. Can you introduce a tag that has the opposite meaning of `has-special-compilation-requirements`?
Like `simple-executable`. We can attach to all the sources that do not have `has-special-compilation-requirements` and have `revng-qa.compiled` to only use `simple-executable`.

Basically `revng-qa.compiled` would become "compile regular, simple, executable".

Cvelth: Sounds good to me.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: 1. Done.
2. Would you be against postponing this to after we moved to assembly-less setup, since I'm still unsure about how much all this is going to change as the result? Maybe that will help us come up with a better name (I hate both `printer`-likes and `decoder`-likes), or 🤞, we'll even be able to assimilate this stuff into some other template.

scripts/abi_artifact_generator/end_to_end.md
Cvelth: Done.

share/revng/test/configuration/revng-qa/for-detect-abi.yml
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
aleclearmind: OK

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
Cvelth: Done.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: Done.

share/revng/test/tests/abi/runner.c
Cvelth: Done.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/tests/abi/runner.c
Cvelth: Done.

share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml
```diff
@@ -0,0 +1,69 @@
...
+  - type: revng-qa.compiled-with-debug-info
+    from:
+      - type: revng-qa.compiled-with-debug-info-copy
+      - type: source
+        filter: abi-test-runner
+      - type: source
+        filter: abi-test-printers
+    command: |-
+      GENERATED_HEADERS=$$(temp -d);
+      ${TRIPLE}objdump --wide --no-show-raw-insn ${OBJDUMP_FLAGS} -h -t -d ${INPUT1}
```
Cvelth: (since the topic has no thread (we discussed it in Mattermost), I'll open one. cc: @aleclearmind)

You asked for simplification of the structure of these `command`s, what do you think about something like this?

aleclearmind: Still way way too complex, several points I mentioned are not addressed.

1. `test_list.txt` can go too. All of its content is deterministic (see [this comment](https://github.com/revng/revng-qa/pull/17#discussion_r1354348110)).
2. Reorganize all the files as follows:
   1. Only one entry in `sources`, targeting a *directory*, `share/revng/test/tests/abi/`. We never did this, if it doesn't work, let's fix `revng-test-configure` so that it works.
        This entry must have `repeat-for:` all the ABIs.
   2. Only *one* command that takes *one* input, the source above and does *everything*. From the only input you can compute the path to all the source files you want.

   This should make go away many tags.

I'd like to stress that commands whose results are used as input in another command are not OK.

Cvelth: I left the `simple-executable` one intact, if that's ok with you. Other than that, please take a look.

share/revng/test/tests/abi/prepare_for_runtime.sh
```diff
@@ -0,0 +1,49 @@
...
+OBJDUMP_FLAGS="$5"
+
+test -n "${INPUT_DIRECTORY}"
+test -n "${OUTPUT_DIRECTORY}"
+test -n "${TRIPLE}"
+test -n "${CFLAGS}"
+
+mkdir -p ${OUTPUT_DIRECTORY}
+
+# Link the `mmap`'ed binary
```
Cvelth: @aleclearmind I went with a separate script for the "combined" runtime-abi-test toolchain - since it's a lot easier to deal with, please review.

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(revng-qa NONE)
 install(PROGRAMS revng-test-configure boolparser.py common.sh
         DESTINATION libexec/revng)
 
-install(DIRECTORY share/ DESTINATION share/)
+install(DIRECTORY share/ DESTINATION share/ USE_SOURCE_PERMISSIONS)
```
Cvelth: @aleclearmind On the topic of scripts, is this the right way to enable them in the `share` directory. This is what `revng`-side does, but I'm not sure if just this is enough.

revng-test-configure
```diff
@@ -297,14 +297,18 @@ class Configuration:
                                       for tag
                                       in self.tags(repetition)
                                       if tag not in tags]
+                derived_targets_prefix = os.path.splitext(path)[0]
```
Cvelth, line 300: @aleclearmind `revng-test-configure` worked well with directories :tada:

This is just a small cosmetic change to convert paths like
```bash
$ORC_ROOT/build/test/revng-qa/default/share/revng/test/tests/abi/-arm-aapcs-abi-revng-qa.runtime-abi-tests-31e2d87e/
```
into
```bash
$ORC_ROOT/build/test/revng-qa/default/share/revng/test/tests/abi/arm-aapcs-abi-revng-qa.runtime-abi-tests-31e2d87e/
```

Please review

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+
+  - name: arm-aapcs-abi
+    implies: [arm]
+    variables:
+      ABI_NAME: AAPCS
+      CFLAGS:
+        - -DABI_AAPCS
+
+sources:
+  - tags: [abi-test-function-prototypes, simple-executable]
```
Cvelth: @aleclearmind I think I mentioned it somewhere, but just to stress, I'm keeping the `simple-executable` on intact, if that's alright with you.

aleclearmind: `functions.c` on its own seems to do close to nothing.
Why do we want to build it through the `simple-executable` toolchain?

If there's a reason add a comment, if there's no strong reason, let's drop it.

CMakeLists.txt
aleclearmind: Works.

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
Cvelth: This is the one we use to extract dwarf information from. I'd rather keep it as close as possible to `calc` and friends (binaries we actively test) to limit any possibilities of it being the reason for a failure or something.

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
```diff
@@ -0,0 +1,109 @@
+tags:
```
aleclearmind: Merge this file with `share/revng/test/configuration/revng-qa/runtime-abi-test-toolchain.yml`

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
aleclearmind: That just saves from typing `${TRIPLE}gcc $INPUT -o $OUTPUT $CFLAGS`, which basically we already do in `prepare_for_runtime.sh`.
On the other hand, it makes `test-conversion.sh` depend on stuff that's not produced by `prepare_for_runtime.sh`, increasing complexity.

`prepare_for_runtime.sh` produces a directory, `test-conversion.sh` consumes it. That's it.

share/revng/test/tests/abi/prepare_for_runtime.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+set -euo pipefail
+
+
```
aleclearmind:
```bash

if test $ARG_COUNT -ne 47; then
  echo "$0 INPUT_DIRECTORY OUTPUT_DIRECTORY ..." > /dev/stderr
  exit 1
fi
```

aleclearmind (changes requested):

share/revng/test/tests/abi/prepare_for_runtime.sh
aleclearmind, line 1: Can we rename this file to something like `build-abi-artifacts.sh` and have in `revng/revng` something like `verify-abi-artifacts.sh`?

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
Cvelth: It's less about the current state, and more about "what if the default gcc invocation changes in the future". Also consider that this will also be build with `msvc` too sooner or later.

But, I guess I'm not against pushing the remaining binaries into the same directory.

TL;DR: I kind of just convinced myself (sorry for the noise), let's indeed move them all together.

share/revng/test/tests/abi/prepare_for_runtime.sh
Cvelth: I kind of hate using `artifact`, now that pipeline uses it, but that is indeed better than the current naming.

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
aleclearmind: The goal of `revng-test-configure` is to easily build pipelines of commands, repeat them, configure them with tags and the like.

If we need to factor code out, we can use support scripts.
Let's not use the pipeline feature for factoring code, it's not its purpose.

share/revng/test/tests/abi/prepare_for_runtime.sh
aleclearmind: Simpler: `build-abi-test-programs.sh` and `verify-abi-test-programs.sh`.

Cvelth: `programs`->`binaries` for consistency?

aleclearmind: Works

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
Cvelth: Done.

share/revng/test/tests/abi/prepare_for_runtime.sh
Cvelth: Done.

share/revng/test/tests/abi/prepare_for_runtime.sh
Cvelth: Done for the `revng-qa`-side script.

As for the `revng`-side, I totally forgot that there are two (and that I quite fond of the names):
- `test-layout.sh` - the stripped down version that is going to stay open-source.
- `test-conversion.sh` - the full version that will eventually move together with `raw -> cabi` converter.

share/revng/test/configuration/revng-qa/runtime-abi-tests.yml
Cvelth: Done.

share/revng/test/tests/abi/prepare_for_runtime.sh
aleclearmind: > going to stay open-source

Are you think to the plan to move `ConvertToCABI` to revng-c?

In any case, can we merge `test-{layout,conversion}.sh`? We can just use `if`s or think about it when we'll actually split and leave a couple of TODOs.
For now, they just seem to be code duplication for me. I'd just collapse them in a single test in `runtime-abi-tests.yml`.
I understand the reasons why we did this, but I'd really like to simplify all of this as much as we can.

Cvelth: I definitely remember that conversation popping up at some point. Either way, merging them back up works for now.

Cvelth: But do you really prefer `verify-abi-test-programs.sh` to something like `test-conversion.sh` or even better `test-function-type-conversion.sh`?

aleclearmind: > I definitely remember that conversation popping up at some point. Either way, merging them back up works for now.

Duplicating code is just confusing, if we want double functionality we can add a parameter.
If you have some previous discussion, feel free to refer to it so we can get a better idea around the reasoning.

> But do you really prefer verify-abi-test-programs.sh to something like test-conversion.sh or even better test-function-type-conversion.sh?

I want something that makes it extremely obvious that one script produces stuff processed by the other script.
That's why I suggested `build-abi-test-programs.sh` and `verify-abi-test-programs.sh`.

Cvelth: Done.

aleclearmind:
Merged in ba440c8. Thanks.

PR #20, opened 2023-08-01
Add msvc support to the abi testing setup

Cvelth:


scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -1,98 +1,126 @@
 ---
 "x86_64":
   register_type: uint64_t
+  intel_register_type: qword
```
aleclearmind: You can deduce this from `register_size`.

scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -1,98 +1,126 @@
...
   register_size: 8
   register_list: [rax, rbx, rcx, rdx, rsi, rdi, r8, r9, r10, r11, r12, r13, r14, r15]
+  styles:
```
aleclearmind: 1. `comment_styles`?
2. Is this the only relevant difference so far between the GNU and the MSVC assembler?

scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -1,98 +1,126 @@
...
   templates:
     call_a_function: |
       call {{ function_name }}
     return_from_function: |
       ret
-    return_from_function_with_cleanup: |
-      ret $8
     fill_a_register: |
-      mov {{ location }} + {{ offset }}, %{{ register }}
+      mov {{ register }}, qword ptr [{{ location }} + {{ offset }}]
```
aleclearmind: INTEL SYNTAX \o/

scripts/abi_artifact_generator/config/architectures.yml
```diff
@@ -1,98 +1,126 @@
...
+      pop {{ register }}
+      mov qword ptr [{{ location }}], {{ register }}
     restore_return_address: |
-      push {{ stack_size }} (%rsp)
+      mov {{ register }}, qword ptr [{{ location }}]
+      push {{ register }}
+    shift_last_stack_word: |
+      pop {{ register }}
+      add rsp, 8
+      push {{ register }}
```
aleclearmind: 1. Why?
2. Unused?
3. Is there documentation for these things?

scripts/abi_artifact_generator/templates/functions.h.jinja_template
```diff
@@ -48,12 +59,25 @@ typedef struct {
 
 /*{ endfor -}*/
 
+#pragma pack(push, 1)
```
aleclearmind: > clang supports the Microsoft #pragma pack feature for controlling record layout. GCC also contains support for this feature, however where MSVC and GCC are incompatible clang follows the MSVC definition.
> -- https://releases.llvm.org/3.4/tools/clang/docs/UsersManual.html

Are you aware of the incompatibility? Does it affect us?

scripts/abi_artifact_generator/templates/functions.h.jinja_template
```diff
@@ -17,6 +19,15 @@ This template file is distributed under the MIT License. See LICENSE.md for deta
 #define ABIDEF
 #endif
 
+/*
+ * Define a `NOINLINE` specifier
```
aleclearmind, line 26: Why do we need this? Docs

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -2,36 +2,44 @@
...
+  CurrentOffset = 0;
+
+  /*{- for type in arguments }*/
+  if (sizeof(/*( type )*/) <= ARGUMENT_SIZE_LIMIT(argument_/*( loop.index - 1 )*/) && !IS_PACKED(/*( type )*/)) {
+    *(/*( type )*/ *) ((uint8_t *) TemporaryLocation + CurrentOffset) = argument_/*( loop.index - 1 )*/;
+    CurrentOffset += sizeof(/*( type )*/);
+  } else {
+    *(/*( type )*/ **) ((uint8_t *) TemporaryLocation + CurrentOffset) = &argument_/*( loop.index - 1 )*/;
+    CurrentOffset += sizeof(/*( type )*/ *);
+  }
```
aleclearmind: I cannot read this.

1. Thoroughly describe what's going on here.
2. Can you reduce the amount of `/* ... */` per line by introducing some temporary variables and local `typedefs`? The code is really really hard to read. `typedefs` lifetime is limited to the current scope, so you can "redefine them" multiple times with the same name (in different scopes).

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -40,61 +48,24 @@ void __attribute__((noinline, weak)) ABIDEF /*( name )*/(
 /* Helps avoiding compiler warnings */
 void *typeless_setup = (void *) set_return_value_up;
 
-#ifdef __SIZEOF_INT128__
-#define INT128_IS_SCALAR 1
-#else
-#define INT128_IS_SCALAR 0
+#ifdef _MSC_VER
+#pragma warning(disable: 4715)
```
aleclearmind: What is this? Docs

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  puts("");
+}
+
+/*{ endfor -}*/
+
+uint8_t TemporaryData[/*( generated_byte_count )*/];
+void *TemporaryLocation = &TemporaryData;
+ptrdiff_t CurrentOffset = 0;
```
aleclearmind: Docs

share/revng/test/tests/abi/functions.c
```diff
@@ -27,3 +29,7 @@ int main(void) {
 
   return 0;
 };
+
+uint8_t TemporaryData[/* doesn't matter */ 1];
+void *TemporaryLocation = &TemporaryData;
+ptrdiff_t CurrentOffset = 0;
```
aleclearmind: Uh?

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -40,61 +48,24 @@ void __attribute__((noinline, weak)) ABIDEF /*( name )*/(
...
     } else {
-      /*( return_value_type )*/ *return_value_pointer =
-          ((/*(- return_value_type )*/ *ABIDEF(*)(void)) typeless_setup)();
-
-      puts("    Function: \"/*( name )*/\"");
-      puts("    Arguments:");
-      puts("    ReturnValue:");
-      POINTER(/*(- return_value_type )*/, &return_value_pointer);
+      /*( return_value_type )*/ *return_value_pointer = ((/*(- return_value_type )*/ *(ABIDEF *)(void)) typeless_setup)();
+      *(/*( return_value_type )*/ **) TemporaryLocation = return_value_pointer;
```
aleclearmind: I cannot read this

aleclearmind (changes requested):

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: It's a bit more complex, in the assembly template, "style-specific" lines start from the given style name - then when rendered all the "non-current" style marks gets replaced with provided here comment indicator.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: I still prefer AT&T (I guess because my first (and favorite) assembly is ARM). MASM, on the other hand, really doesn't like it :enterprise:

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Ah, whoops. It's a leftover from a previous iteration. Can be removed now.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: No, I was not aware of it, but I don't think it has any impact in our case. Here all we want to do is to make the compiler generate these structs with unnatural alignment. Which is the case for both, as we don't really care about the internals per-se.

share/revng/test/tests/abi/functions.c
Cvelth: This is for `function` definition binary. It's only ever used as dwarf/pdb source - so the internal don't really matter, as long as it compiles fine, or in this case, links fine.

scripts/abi_artifact_generator/config/architectures.yml
aleclearmind: So when generating code for MSVC we're emitting comments with stuff that would be generated for GCC?
That's an additional hidden layer of processor. Can't we just use an `if` in the template engine?

Cvelth: I really thought about that, but both ways sound about equally complex. Except having the outputs side-by-side made debugging them a lot easier (since you see both renderings of the things at once), kind of like here:
```x86
;; .global execute_single_huge_struct_argument
;; execute_single_huge_struct_argument:
execute_single_huge_struct_argument proc
  {{ big piece of assembly omitted because it's the same for both }}
;; #ifndef CalleeIsResponsibleForStackCleanup
  add esp, 512
;; #endif
  mov eax, dword ptr [saved_return_address]
  push eax
  ret
execute_single_huge_struct_argument endp
```

Even while on msvc version you can still see the equivalent gcc code.

Cvelth: But I guess if you'd rather the python-side harbored the extra complexity, I'm not too opposed.

aleclearmind:
```

/*( gcc )*/#include "../abi.h"

/*( gcc )*/#if defined(TARGET_x86_64) || defined(TARGET_i386)
/*( gcc )*/.intel_syntax noprefix
/*( gcc )*/#endif

/*( msvc )*/ifndef rax
/*( msvc )*/.model flat, c
/*( msvc )*/endif

/*( msvc )*/extern saved_return_address: /*( intel_register_type )*/
/*( msvc )*/extern randomized_state: /*( intel_register_type )*/

/*( msvc )*/extern regenerate_state: proc
/*( msvc )*/extern print_current_state: proc
```

vs

```
/* if gcc */
#include "../abi.h"

#if defined(TARGET_x86_64) || defined(TARGET_i386)
.intel_syntax noprefix
#endif
/* elif msvc: */
ifndef rax
.model flat, c
endif

extern saved_return_address: /*( intel_register_type )*/
extern randomized_state: /*( intel_register_type )*/

extern regenerate_state: proc
extern print_current_state: proc
/* endif */
```

Second seems much better to me, plus no polluted output.
It's more debatable on single line situations, but still, let's go with the `if`s.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
aleclearmind: OK, maybe leave a comment about this.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Sounds good.

scripts/abi_artifact_generator/config/architectures.yml
aleclearmind: Ping

scripts/abi_artifact_generator/config/architectures.yml
aleclearmind: Updates?

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: I'm pretty sure I did this. Looks like this is not the latest version (I pushed a wrong thing), sorry.

Cvelth: Pushed it.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Pushed it. Not sure if I prefer the new approach from readability standpoint though

aleclearmind:
Converting this to draft since we first want to merge [feature/i386-microsoft-abis](https://github.com/revng/revng/pull/280).

Cvelth:
(this thread is a reminder about a carry-over thread from #17):
https://github.com/revng/revng-qa/pull/17#discussion_r1354321365

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,199 @@
...
+#*/
+
+/*(- generation_notice )*/
+
+#include "../common.h"
+#include "../functions.h"
+#include "../types.h"
+
+_Static_assert(sizeof(uint8_t) == 1, "A type with size == 1 is required.");
+#define MAKE_PRINT_HELPER(TYPE, POINTER, RESULT) \
```
Cvelth: This thread is a reminder about a carry-over thread:
https://github.com/revng/revng-qa/pull/17#discussion_r1354322243

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
```diff
@@ -0,0 +1,199 @@
+/*#-
+This template file is distributed under the MIT License. See LICENSE.md for details.
+#*/
+
+/*(- generation_notice )*/
```
Cvelth: This thread is a reminder about a carry-over thread:
https://github.com/revng/revng-qa/pull/17#discussion_r1354321365

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: Closing all the documentation-related threads for now.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: Closing all the documentation-related threads for now.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: Closing all the documentation-related threads for now.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: Closing all the documentation-related threads for now.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: Closing all the documentation-related threads for now.

scripts/abi_artifact_generator/templates/functions.h.jinja_template
Cvelth: This will go in as a part of #17.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: Done.

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: This will go in as a part of #17

scripts/abi_artifact_generator/config/architectures.yml
Cvelth: This will go in as a part of #17.

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: > Hoist in one of the headers above or in a new one?
Same with functions at the bottom.

This was streamlined somewhat (not as many macros :tada:)

But still, consider that this is only used here, you sure you want a header for it?

scripts/abi_artifact_generator/templates/printers.inc.jinja_template
Cvelth: > AFAIU this decodes a global buffer containing data dumped by the various functions. `printers.c` does not convey this very well. Can we give a proper name to the buffer above and call all of this `decode_$NAME.c`?

Done. Also, I split the buffer up, now it's `value_dumps`, `size_dumps` and `address_dumps`. What do you think?

share/revng/test/configuration/revng-qa/architectures.yml
```diff
@@ -30,47 +30,48 @@ tags:
...
+      COMMON_FLAGS:
         - -DTARGET_arm
   - name: x86-64
     implies: [cross-compiler]
     variables:
       TRIPLE: x86_64-gentoo-linux-musl-
       QEMU_NAME: x86_64
       OBJDUMP_FLAGS: -Mintel
-      CFLAGS:
+      COMMON_FLAGS:
```
aleclearmind: Can you better describe what's the difference between `CFLAGS` and `COMMON_FLAGS`?

AFAIU the former are specific to GNU-compatible drivers (i.e., `gcc` and `clang`) while the others work for `cl.exe` too, is that right?

If so, I propose `CFLAGS` -> `GCC_CFLAGS` and `COMMON` -> `COMMON_CFLAGS`.

share/revng/test/configuration/revng-qa/architectures.yml
```diff
@@ -1,5 +1,15 @@
...
   - name: linux
+  - name: windows
+    variables:
+      MSVC_CFLAGS:
+        - -WX
+        - -nologo
+      MSVC_LDFLAGS:
+        - /debug
+        - /nologo
+        - /wx
```
aleclearmind: 1. Let's be consistent with the prefix for flags. Given that MSVC supports `-` as a prefix and if we use it we can reused `-DTARGET_x86`, I'd stick with it.
2. Add a comment for the reason for each option. The average reader won't be familiar with them.

share/revng/test/configuration/revng-qa/for-import-pe.yml
```diff
@@ -3,19 +3,46 @@
...
+        - -Z7
+        - -EHsc
+        - -O2
+  - name: vc16-i386
+    variables:
+      MSVC_TRIPLE: "i386-winsdk-vc16-"
+      MSVC_CFLAGS:
+        - -Z7
+        - -EHsc
+        - -O2
```
aleclearmind: 1. Consider factoring `["-Z7", "-EHsc", "-O2"]`  into a new tag that's implied by `vc19-x86-64`, `vc19-i386` and `vc16-i386`-
2. Comment meaning of flags

share/revng/test/configuration/revng-qa/for-import-pe.yml
```diff
@@ -26,15 +53,25 @@ commands:
...
+      timeout 30 ${MSVC_TRIPLE}cl z:\$INPUT $MSVC_CFLAGS $COMMON_FLAGS -link /pdb:$OUTPUT $MSVC_LDFLAGS
+
+#   - type: revng-qa.disassembled
+#     from:
+#       - type: revng-qa.compiled
+#         filter: windows
+#     suffix: .asm
+#     pool: wine
+#     command: |-
+#       timeout 60 ${MSVC_TRIPLE}dumpbin -nologo -disasm:nobytes $INPUT -out:$OUTPUT
```
aleclearmind: What do we want to do with this?

share/revng/test/configuration/revng-qa/for-import-pe.yml
```diff
@@ -3,19 +3,46 @@
...
 sources:
-  - tags: [windows]
+  - tags: [windows, simple-executable]
+    repeat-for:
+      - [vc19-x86-64, msvc-dynamic-linkage]
+      - [vc19-i386, msvc-dynamic-linkage]
+      - [vc16-i386, msvc-dynamic-linkage]
+      - [vc19-x86-64, msvc-static-linkage]
+      - [vc19-i386, msvc-static-linkage]
+      - [vc16-i386, msvc-static-linkage]
```
aleclearmind: Why do we care about static linking?
Doing static linking in windows is a suspicious thing.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -62,3 +62,28 @@ NOINLINE /*( return_value_type )*/ ABIDEF test_/*( name )*/(void) {
...
+      /*{- endfor -}*/
+    );
+  }
+/*{- endfor }*/
+
+/*{ for name in return_value_functions -}*/
+  test_/*( name )*/();
+/*{ endfor }*/
+}
+#endif
```
aleclearmind: So what's the final decision on `-O0` vs `-O2`?
I suspect this wouldn't be necessary if we stick with `-O0`.

An alternative to call the functions is to store their address in a global variable, I think.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -16,6 +16,15 @@ tags:
       MSVC_CFLAGS:
         - -DABIDEF=
 
+  - name: x86-64-vectorcall-microsoft-abi
+    implies: [x86-64]
+    variables:
+      ABI_NAME: Microsoft_x86_64_vectorcall
+      CFLAGS:
+        - -DABIDEF=UNSUPPORTED # safeguard in case somebody tries using gcc
```
aleclearmind: Put the comments on their own line

revng-test-configure
```diff
@@ -76,22 +76,23 @@ def most_similar_tuples(inputs, command_type):
...
-                error(f"While emitting {command_type}, {reference.path} has multiple good matches for index #{index}:\n  {inputs[index][scores[0][2]].path}\n  {inputs[index][scores[1][2]].path}")
+                error(f"While emitting {command_type}, {reference.path} has multiple good matches for index #{index}:\n  {inputs[index][scores[0][3]].path}\n  {inputs[index][scores[1][3]].path}")
 
             if not scores:
                 error(f"While emitting {command_type}, cannot match {reference.path} with any other target for input #{index}")
 
-            used_inputs.add((index, scores[0][2]))
-            entry[index] = inputs[index][scores[0][2]]
+            used_inputs.add((index, scores[0][3]))
+            entry[index] = inputs[index][scores[0][3]]
```
aleclearmind: Please thoroughly elaborate. I'm very sensitive to changes in this area.

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -23,67 +23,92 @@ def accept_arguments():
...
-        data,
-    ):
-        result.append((match[1].decode("utf-8"), match[2].decode("utf-8")))
 
-    return result
+def select_parser(sections, symbols, disassembly):
+    if re.match(b"^\n[^:]+:\s+file format elf.+\n", sections):
+        return ObjdumpOutputParser(sections, symbols, disassembly)
+    elif re.match(b"^\r\nDump of file .*\.exe\r\n\r\nPE signature found\r\n", sections):
+        return DumpbinOutputParser(sections, symbols, disassembly)
```
aleclearmind, line 177: Is the support to `dumpbin` new? If so, mention that in the commit message.

share/revng/test/tests/abi/gather_symbols.py
```diff
@@ -99,7 +99,59 @@ def find_callsites(self, name):
 
 class DumpbinOutputParser:
```
aleclearmind: If there's a common base API between the two parser, create a base class inheriting from `ABC` and use `@abstractmethod`: https://docs.python.org/3/library/abc.html#abc.abstractmethod

share/revng/test/tests/abi/build_abi_test_binaries_with_msvc.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
aleclearmind, line 1: NOTETOSELF: review

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -115,14 +115,27 @@ sources:
   - tags: [for-runtime-abi-tests]
     repeat-for:
       - [x86-64-systemv-abi]
+
+      # TODO: re-enable once new PDB support is added
```
aleclearmind: What's this about?

aleclearmind (changes requested):

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: Yep, exactly. there's not much point of duplicating all the `-DSOMETHING` flags since they work the same way on both toolchains.

👌 on `GCC_CFLAGS` and `COMMON_CFLAGS`.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: 1. That's the problem, linker doesn't. :msvc_burn:
2.1. I'd imagine `/debug` and `/nologo` are pretty self-explanatory. But yeah, I'll comment on the other ones.
2.2. I didn't add these. I just moved them a bit up from `import-pe` :)

aleclearmind:
Can you better describe what the "Rework decoding to accept a full state " commit does?

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: Ah, I thought I removed it. It's the mirror of the same command for gcc toolchain. I didn't find any users of that one - so I assumed it's a debug helper. So we might want to have the same things for msvc.

BUT it's way too slow. `dumpbin` is not the fastest utility as it is, and wine is not doing it any favours either.

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: I'm doing static linkage as the part `mmap`'ed binary. And this is just a way of guarding myself from potential issues - if something is wrong with the way we treat unusual statically-linked windows binaries - i'd rather this test also shown it and not only my stuff.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: 1. Still hesitating on it. Once I wrapped up the bugfixes I have open now (since I'm almost done with that PR), I was planning to look deeper into it.

2. That should help with the tail-call workaround. Not so sure about this one though. Since, the problem here is the msvc linker. And we kind of need `/opt:ref` otherwise the binary is really big (filled with a whole lot of stuff we don't care about) which makes all the stuff we do with it really slow. And `/opt:ref` also happens to remove all these functions unless they are reachable from the main.

> An alternative to call the functions is to store their address in a global variable

Actually, that's a really cool idea. If that works, i'd greatly prefer it.

revng-test-configure
Cvelth: Ah, I was planning to do a more explicit commit message for this but I guess I just forgot.

The gist is, if there is a situation where a command has two inputs:
* `input1` has one possibility:
  * tags `A`, `B`, `C`
* `input2` has two possibilities:
  * tags `A`, `B`, `D`
  * tags `A`, `B`, `E`, `F`

The current `develop` approach would consider the scores for these two equal (both have two matches) - so it errors out.
My change makes it so it's also inversely proportional to the difference between the tags.

So, in this example, the `input2-1` would get score `[false, -2, 2]`, and `input2-2` - score `[false, -2, 3]` - making the first one a clear match.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: There was a `TODO` in the file. But sure thing.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: Because we don't handle PDBs this new correctly yet, these tests just fail without doing anything useful. So I disabled them for now. I left a bug-report in `~model-import`

scripts/abi_artifact_generator/templates/decoders.inc.jinja_template
```diff
@@ -20,103 +20,88 @@ This template file is distributed under the MIT License. See LICENSE.md for deta
...
-      POINTER(TYPE, (LOCATION) + (OFFSET));                                    \
-      (OFFSET) += sizeof(TYPE *);                                              \
-    }                                                                          \
-  } while (0)
-
-ptrdiff_t current_offset = 0;
 /*{ for name, arguments in argument_functions.items() -}*/
-static void ABIDEF decode_/*( name )*/(const uint8_t *output) {
-  current_offset = 0;
+static void ABIDEF decode_/*( name )*/(const struct encoded value) {
```
Cvelth: > Can you better describe what the `"Rework decoding to accept a full state"` commit does?

(cc: @aleclearmind)

I guess the name sucks indeed. The idea is to rework this interface to accept something a bit more structured as opposed to a stream of bytes. As in, to give a proper-ish interface to this header.

share/revng/test/configuration/revng-qa/architectures.yml
aleclearmind: 1. fml. We could have a wrapper do it... But OK. Add a comment.
2. 2. Be a good citizen and improve what you see `:)`.

share/revng/test/configuration/revng-qa/for-import-pe.yml
aleclearmind: `llvm-objdump`?

share/revng/test/configuration/revng-qa/for-import-pe.yml
aleclearmind: Mmmmmh, I don't understand. Can you detail a scenario that this change prevents/makes easier to detect?

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
aleclearmind: 2. > which makes all the stuff we do with it really slow

   How can the linker make us slower? I'm a bit suspicious. I would understand if, by default, there's a lot of code and `/opt:ref` DCE's it away, but otherwise I don't see how a "bigger binary" can make us slower. Is there more code? Just padding?

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: I don't think revng-qa depends on llvm, and I'm not sure it's worth it just for this? Hell, I'm not even sure we need this thing in the first place.

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: Since these end up in the `revng-qa.compiled` pile, they go through the usual pipeline, don't they? So in case something within revng breaks and statically linked windows binaries no longer work, these tests would be the first to fire.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: Don't forget that the binary is linked statically. MSVC toolchain pulls a whole lot of libraries in. `/opt:ref` eliminates practically all of them, leaving us with pretty lean models.

share/revng/test/configuration/revng-qa/for-import-pe.yml
aleclearmind: We'd need to depend on `clang-release` which is fine.
I'd be happy to have it, but as you say is not strictly necessary.
Up to you!

share/revng/test/configuration/revng-qa/for-import-pe.yml
aleclearmind: So this is not strictly related to ABI stuff? Interesting.
I've never seen a statically linked binary, can you send me one?

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
aleclearmind: Why link statically?

Cvelth: Because we call functions from said binary from gcc. It's definitely a lot safer (for a certain definition of safer) for the binary to be self-contained and to not try to dynamically link to anything.

share/revng/test/configuration/revng-qa/for-import-pe.yml
aleclearmind: OK. Static linking is not people actually do on Windows, since the compatibility layer offered by the OS are the system DLLs, not the interface with the kernel as in Linux.

So, if this slows things down too much, or somehow forces us to force linker optimizations in completely unrelated parts of the testing suite (the ABI part), I'd drop it.

It's nice to have, but not a real world use case. I don't want to slow down the test suite or make it more complex in order to have "nice to have" things. Let's keep things simple.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
aleclearmind: I suspect this is FUD: we already rely on the fact that the functions we call do not call anyone else.

Let's avoid complications, and use regular dynamic binaries unless you have proof that we need statically linked binaries.

revng-test-configure
aleclearmind: Provide a concrete example in the codebase where not having this is a problem.
I'm striving to really *not have* to rely on more complex features in there and find alternative solutions.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -105,12 +144,37 @@ commands:
...
+#        exit 1;
+#      fi;
+#      "${INPUT}/build_abi_test_binaries_with_msvc.sh"
+#        "${INPUT}"
+#        "${OUTPUT}"
+#        "${TRIPLE}"
+#        "$${MSVC_TRIPLE}"
+#        "${CFLAGS} ${COMMON_FLAGS}"
+#        "${MSVC_CFLAGS} ${COMMON_FLAGS}"
+#        "${MSVC_LDFLAGS}";
```
aleclearmind: 1. Can we use `TRIPLE` instead of `QEMU_NAME`?
2. Can we push the computation of `MSVC_TRIPLE` into the script?
3. I'm not sure about the state of this. I don't want to merge unused code (`build_abi_test_binaries_with_msvc.sh`).
   1. Is your plan to wait for the PDB bug to be fixed before merging this?
   2. AFAIU we need newer versions of MSVC in order to test this. So how have you been working so far, given that newer PDBs are broken?
   3. Can we start to merge some stuff using only older PDB format?

aleclearmind (changes requested):

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: 1-2. 👌

3. I really want to flush this. I have been working on this for way too long and would like to synchronize the states, so to say.

> So how have you been working so far, given that newer PDBs are broken?

Even if importing is broken, it's not hard to "emulate it": make models for interesting functions manually and then run the tests with those. For obvious reason, I only did a couple, but the artifacts look good, now we only need the non-manual importing to work for everything to click.

Either of the options work for me:
1. Remove only the `for-runtime-abi-tests.yml` part for now - reintroduce it once PDBs are supported.
2. Make this produce the artifacts (that I already verified to be good), but add a tag for revng-side to ignore them until PDBs are fixed.
3. Try to see if it would work only with `vc16`, considering we no longer need assembly, but I'm skeptical about this.
4. Try to look if there's a way to run `vc19` in some kind of compatibility mode and produce older PDBs (doubt it).

My preferred option is 2, I guess.

aleclearmind: Let's give a couple more days to @dwarfyresearch to look into this and let's see.
In the meantime, no tags for unused things. Just leave the production of the artifacts commented explaining when this can be un-commented.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: Done.

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: Done.

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: Done.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: Done.

share/revng/test/configuration/revng-qa/for-import-pe.yml
Cvelth: These are gone.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: Done.

revng-test-configure
Cvelth: I _really_ should start double-checking these before asking for reviews. It looks like some unrelated change hid this issue. So I'm closing this thread. Will reopen if I run into it again.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: I moved these lines to the next (`dumpbin`-focused) commit. Closing the thread.

share/revng/test/tests/abi/gather_symbols.py
Cvelth: Done.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
aleclearmind: Status on this? Did the @dwarfyresearch patch fixes things?

Cvelth: Yep, the patch looks good. But it unearthed some problems I somehow missed before. Shouldn't be hard to fix though.

Also, we should probably prioritize the post-dla-related fixes first. There might be some overlap, and that PR is practically ready to go in.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
aleclearmind: Ping?

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
aleclearmind: Status?

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: > So what's the final decision on -O0 vs -O2?

I went with `pragma optimize off` around the offending part.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: Closing this since these are re-enabled.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: Closing this thread since this code is uncommented again.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
aleclearmind: OK, please add a comment about why we need this.

Are there any other places where we need to do tricks to fight the compiler trying to optimize the code away?
IIRC the argument in favor of `-O2` over `-O0` is readability of the code, but let's make sure we don't need 10 hacks around the codebase to keep `-O2`.

Cvelth: It's specifically the tail call optimization. And even then - it's only because of the way we inject break-points. We could, _theoretically_ rework that part instead - with support for tail calls, but, let's be real, it's a lot easier to just force compiler not to optimize those.

Also, it's only `-O0` on MSVC, on GCC it's `#pragma GCC optimize ("no-optimize-sibling-calls")`.

Cvelth: IMO, I'd still rather go with the "fake-call" (I also tried various other ideas you had, like writing the address of the function to a global variable - they all still cause trouble to varying degrees) to prevent the optimization, but this works.

aleclearmind: > Also, it's only -O0 on MSVC, on GCC it's #pragma GCC optimize ("no-optimize-sibling-calls").

I'm not concerned about the fact that we disable optimizations. I'm concerned about the fact that we *enable* them.

So my question is: are there other places that would become significantly simpler if we were to globally switch from `-O2` to `-O0`?

Cvelth: > are there other places that would become significantly simpler if we were to globally switch from `-O2` to `-O0`?

Not that I know of. Aside from pretty obvious tail-call related trap (we never hit the breakpoint we set at the instruction right after the `call` one), I haven't seen anything bad happen because of optimizations.

> I'm concerned about the fact that we _enable_ them.

I feel the opposite way. The fact that even something intrusive as compiler optimizing stuff does not break any of our assumptions makes me feel like we're doing something right. Bonus point: the code we're testing is a lot closer to the code we're going to see in the real world too.

scripts/abi_artifact_generator/config/functions.yml
```diff
@@ -10,7 +10,7 @@ structs:
   one_byte_struct: ["uint8_t a"]
   two_byte_struct: ["uint16_t a"]
   small_struct_of_structs: ["two_byte_struct a", "one_byte_struct b"]
-  struct_of_structs: ["small_struct a[2]", "one_byte_struct b"]
```
aleclearmind, line 13: >  (drop if possible) Avoid nested struct array since import-pe doesn't support them

Fixed?

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
```diff
@@ -57,6 +57,11 @@ NOINLINE /*( return_value_type )*/ ABIDEF test_/*( name )*/(void) {
   _Static_assert(sizeof(/*( return_value_type )*/) <= /*( maximum_argument_size )*/,
                  "Size of the return value exceeds the allowed maximum.");
   /*( return_value_type )*/ returned = *(/*( return_value_type )*/ *) expected_state;
+
+#ifdef _MSC_VER
```
aleclearmind, line 63: Why only for MSVC?
Comment.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -68,28 +68,61 @@ tags:
     variables:
       ABI_NAME: Microsoft_x86_cdecl
       GCC_CFLAGS:
-        - -DABIDEF=__attribute__((ms_abi,cdecl))
+        # safeguard in case somebody tries using it with gcc
```
aleclearmind: 1. `s/safeguard/Safeguard/`
2. Maybe we could factor these out in a new tag `disable-for-gcc`? Probably not worth it, just an idea.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -68,28 +68,61 @@ tags:
...
+        - -DABIDEF=UNSUPPORTED
       MSVC_CFLAGS:
         - -DABIDEF=__cdecl
 
+  - name: i386-cdecl-microsoft-abi-gcc-edition
```
aleclearmind, line 76: Why do we test GCC at all given that we have MSVC?

Given how badly GCC implements MS ABIs, it feels like a waste of time.
Let's try to save on test runtimes, unless there's a real benefit.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
```diff
@@ -29,6 +29,10 @@ mkdir -p "${OUTPUT_DIRECTORY}"
   ${CFLAGS} -O3 -ggdb3 \
   "${INPUT_DIRECTORY}/functions.c" \
   -o "${OUTPUT_DIRECTORY}/functions"
+"${TRIPLE}gcc" \
+  ${CFLAGS} -O3 \
+  "${INPUT_DIRECTORY}/functions.c" \
+  -o "${OUTPUT_DIRECTORY}/functions-no-debug"
```
aleclearmind: > It doesn't test anything but the fact that we don't crash during the conversion

I'm afraid we're making the time of the test suite explode.
How many binaries are we sending down the decompilation pipeline? Is there a real benefit in testing this?

We're supposed to test the ABI here, not have a bunch of diverse binaries for the sake of it.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: 1. Done.
2. That would feel weird. Especially considering that we also have those we don't want to put through MSVC.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: > Why do we test GCC at all given that we have MSVC?

64-bit one is unusably bad. So I wouldn't even think about trying to put that one through, but 32-bit (even if incompatible) are kind of usable. And it doesn't require any extra effort to support those (they don't do anything weird or unusual - kind of the same as `regpart` ABIs but with msvc-like registers used).

Either way, I'm fine with either keeping them, removing or, maybe, just commenting them out. That way we can still preserve them if we ever run into related problems (I'd imagine there is at least some code out there in the wild compiled by GCC in "msvc-abi" mode, hell, MinGW exists), but we don't have to run it every time we run the tests.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: Actually, I just forgot to remove this (I only removed revng-counterpart).

It doesn't work. MSVC ones assert in `DetectABI`.

The whole point I even added this was because I was experiencing similar problems when doing post-dla-conversion analysis. So having this as a reliable way to reproduce problems helped with debugging a lot.

TL;DR: I didn't mean to push this in, my bad.

scripts/abi_artifact_generator/config/functions.yml
Cvelth: Yep.

scripts/abi_artifact_generator/templates/functions.inc.jinja_template
Cvelth: Turns out, different compilers like different thing.

From what I gathered, for MSVC it's better to dump return value address _before_ it's been returned, and _after_ for GCC, hence this pair of `#ifdef _MSC_VER` here and `#ifndef _MSC_VER` in `setup.inc`.

Cvelth: (comment is done)

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
aleclearmind: OK to comment them and leave a comment on why the might be useful.
Let's strive not to increase the test duration unless there's a tangible benefit.

If we ever want to support mingw binaries, we'd have better implementing a new ABI that more closely describes what mingw actually does, which seems to be radically different from what MSVC does (for 64-bits).

aleclearmind: Is this done?

Cvelth: Yep, I just forgot to close the thread.

https://github.com/revng/revng-qa/pull/20/commits/4d89a0e59160b75fe2cbdb0b0bd7042b6b669ed6#diff-be7a1dce369aaf0e0d90ec71e4cca629675ac7af9b0c1d77f8e7d65509e850d2R167

aleclearmind:
Merged in 46f2ab9. Thanks.

PR #21, opened 2024-02-28
`switch` value simplification

dwarfyresearch:


share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S
```diff
@@ -10,12 +10,16 @@ _start:
...
+    mov rax, 0x3
     ret
 two:
+    mov rax, 0x5
     ret
 three:
+    mov rax, 0x6
     ret
 end:
+    mov rax, 0x1
```
aleclearmind: Are you actually using this test case?
Or are you just using those in `for-simplify-switch.yml`?

share/revng/test/configuration/revng-qa/for-simplify-switch.yml
```diff
@@ -0,0 +1,22 @@
...
+        - -fomit-frame-pointer
+        - -fno-reorder-functions
+        - -fno-unwind-tables
+        - -fno-asynchronous-unwind-tables
+        - -fno-stack-check
+        - -fno-optimize-sibling-calls
+        - -fno-inline-functions
+        - -fno-inline-small-functions
+        - -fno-align-functions
+        - -fno-optimize-sibling-calls
```
aleclearmind: These make no sense here. We're just invoking the assembler, don't blindly copy paste!

share/revng/test/tests/analysis/SimplifySwitch/x86_64/simple-switch.S
```diff
@@ -0,0 +1,100 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+    .intel_syntax noprefix
+    .type myfunc, @function
+    .globl myfunc
+myfunc:
+    mov    rax, QWORD PTR [rax]
```
aleclearmind: 1. Use `rdi`.
2. This function does not make much sense to me, there's a lot of unreachable code.
   Why do you create so much code but then just use 5?
   If you just use 5, why not conflate this test into `simple-switch.S`.

aleclearmind:
1. Rewrite the history using your `@rev.ng` e-mail address
2. Change commit message of `Modify x86_64/switch-jump-table to avoid llvm opts` to
   ```
   x86_64/switch-jump-table.S: add return value

   This prevents LLVM optimizations down the pipeline.
   ```

aleclearmind (changes requested):

share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S
dwarfyresearch: It was a leftover, but we will use this .S file for the test.

share/revng/test/tests/analysis/SimplifySwitch/x86_64/simple-switch.S
dwarfyresearch: We will use the existing `x86_64/switch-jump-table.S` test. Thanks.

share/revng/test/tests/analysis/CollectCFG/x86_64/switch-jump-table.S
```diff
@@ -3,24 +3,45 @@
...
     .intel_syntax noprefix
-    .type _start, @function
-    .globl _start
-_start:
-    cmp    rax,0x3
+    .type myfunc, @function
+    .globl myfunc
+myfunc:
+    mov    rax, QWORD PTR [rax]
+    cmp    rax, 0x5
```
aleclearmind, line 14: Add a detailed comment about why we need this and under what conditions it can be dropped (i.e., `AdvancedValueInfo` does not play well with tracking arguments).

aleclearmind (changes requested):

pfez:
Merged in cde017a4eceedebab1b7cb0c31c3bb96e61eedcd

Thank you!

PR #22, opened 2024-03-13
Add various tests for decompilation

pfez:
This patchset adds tests for various features of the decompiler:
- DLA capability to recover complex data structure like linked-lists and arrays
- DLA capability to update segment and section types, so that we emit nice looking accesses to segments in C
- Capability to emit nice looking integer literals in C
- Capability to emit inline string literals in C

All the actual tests are implmented in `revng-c`, based on `FileCheck`. Here we're adding only the original source, and recipes to compile the code to generated the binaries that are then used in `revng-c` for tests.

### DIFF
--- a/share/revng/test/configuration/revng-qa/for-data-layout-analysis.yml
+++ b/share/revng/test/configuration/revng-qa/for-data-layout-analysis.yml
@@ -1,0 +1,21 @@
+tags:
+  - name: for-data-layout-analysis
+    variables:
+      CFLAGS:
+        - -O2
+        - -Wno-pointer-bool-conversion
+        - -fno-unroll-loops
+        - -fno-inline
+        - -fno-optimize-sibling-calls
+        - -fno-stack-protector
+sources:
+  - tags: [for-data-layout-analysis]
+    members:
+      - share/revng/test/tests/analysis/data-layout-analysis/linked-lists.c
+      - share/revng/test/tests/analysis/data-layout-analysis/segments-and-sections.c
+commands:
+  - type: revng-qa.compiled-for-data-layout-analysis
+    from:
+      - type: source
+        filter: for-data-layout-analysis
+    command: clang $INPUT -o $OUTPUT $CFLAGS
--- a/share/revng/test/configuration/revng-qa/for-pretty-int-formatting.yml
+++ b/share/revng/test/configuration/revng-qa/for-pretty-int-formatting.yml
@@ -1,0 +1,19 @@
+tags:
+  - name: for-pretty-int-formatting
+    variables:
+      CFLAGS:
+        - -O2
+        - -fno-inline
+        - -fno-optimize-sibling-calls
+        - -fno-stack-protector
+sources:
+  - tags: [for-pretty-int-formatting]
+    members:
+      - share/revng/test/tests/analysis/pretty-int-formatting/pretty-ints.c
+commands:
+  - type: revng-qa.compiled-for-pretty-int-formatting
+    from:
+      - type: source
+        filter: for-pretty-int-formatting
+    command: clang $INPUT -o $OUTPUT $CFLAGS
+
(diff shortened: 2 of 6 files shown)


share/revng/test/tests/analysis/data-layout-analysis/linked-lists.c
```diff
@@ -0,0 +1,41 @@
+#include <stddef.h>
```
aleclearmind, line 1: Check conventions

share/revng/test/tests/analysis/data-layout-analysis/linked-lists.c
```diff
@@ -0,0 +1,41 @@
...
+#include <stdlib.h>
+
+#define DATA_LEN 5
+
+struct node {
+  int64_t data[DATA_LEN];
+  struct node *next;
+};
+
+__attribute__((weak)) struct node *init_list() {
```
aleclearmind, line 12:
```c
#define WEAK ...
```

share/revng/test/configuration/revng-qa/for-data-layout-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+  - tags: [for-data-layout-analysis]
+    members:
+      - share/revng/test/tests/analysis/data-layout-analysis/linked-lists.c
+      - share/revng/test/tests/analysis/data-layout-analysis/segments-and-sections.c
+commands:
+  - type: revng-qa.compiled-for-data-layout-analysis
+    from:
+      - type: source
+        filter: for-data-layout-analysis
+    command: clang $INPUT -o $OUTPUT $CFLAGS
```
aleclearmind, line 21: 1. Given that we're not cross-compiling, we could move all of this to `revng-c` directly.

   The reason to put stuff into `revng-qa` is to avoid devs from downloading all the cross-toolchains when running `test/revng` and `test/revng-c`.

   In fact, I'd collapse this into a single command in `test/revng-c` that compiles, decompiles and does whatever it needs to do.
   Same for `for-pretty-int-formatting.yml`.
2. Add a comment on the reasons why we need `clang` and why we want the `CFLAGS` above. We'll forget in 3, 2, 1...

share/revng/test/configuration/revng-qa/for-pretty-int-formatting.yml
```diff
@@ -0,0 +1,19 @@
...
+  - tags: [for-pretty-int-formatting]
+    members:
+      - share/revng/test/tests/analysis/pretty-int-formatting/pretty-ints.c
+commands:
+  - type: revng-qa.compiled-for-pretty-int-formatting
+    from:
+      - type: source
+        filter: for-pretty-int-formatting
+    command: clang $INPUT -o $OUTPUT $CFLAGS
+
```
aleclearmind, line 19: Extra empty line

aleclearmind (changes requested):

share/revng/test/configuration/revng-qa/for-data-layout-analysis.yml
```diff
@@ -0,0 +1,21 @@
...
+        - -Wno-pointer-bool-conversion
+        - -fno-unroll-loops
+        - -fno-inline
+        - -fno-optimize-sibling-calls
+        - -fno-stack-protector
+sources:
+  - tags: [for-data-layout-analysis]
+    members:
+      - share/revng/test/tests/analysis/data-layout-analysis/linked-lists.c
+      - share/revng/test/tests/analysis/data-layout-analysis/segments-and-sections.c
```
aleclearmind, line 15: Use `prefix: share/revng/test/tests/analysis/data-layout-analysis/`. See `for-detect-abi.yml` as an example, in `revng`.

pfez:
Ok I'll take the reviews, and move this to `revng-c`. Thanks for the feedback. Closing the MR here then.

PR #23, opened 2024-03-15
Miscellaneous fixes for Tier 2

pfez:
This commit removes old obsolete decompilation tests, that were no longer used, and replaces them with tests for various features of the decompiler.
* DLA capability to recover complex data structure like linked-lists and arrays.
* DLA capability to update segment and section types, so that we emit nice looking accesses to segments in C.
* Capability to emit nice looking integer literals in C
* Capability to emit inline string literals in C and update the model types of the segments containg such string literals.

All the actual tests are implmented in revng-c, based on FileCheck. Here we're adding only the original source, and recipes to compile the code to generated the binaries that are then used in revng-c for tests.


pfez:
This is the updated version of https://github.com/revng/revng-qa/pull/22

I already integrated all the comments that were opened there

share/revng/test/tests/analysis/Decompilation/x86-64/pretty-ints.c
```diff
@@ -0,0 +1,57 @@
...
+         + arithmetic_shift_right(argc) + bitwise_or(argc) + bitwise_and(argc)
+         + bitwise_xor(argc) + write_and_get_char();
+}
+
+WEAK int argc;
+WEAK const char **argv;
+
+void _start() {
+  main(argc, argv);
+}
```
aleclearmind: Usually you either have `_start` or `main`. Having both is a bit confusing.

Drop `main` and use the global directly?

Same in `segments-and-sections.c`.

aleclearmind (changes requested):

share/revng/test/tests/analysis/Decompilation/x86-64/pretty-ints.c
pfez: `_start` is annoying to use in `FileCheck` because it starts with `_` which is reserved, so it gets the whole `unreserved` prefix.

So it's better to have a wrapper function that you can use in `FileCheck` vanilla, which is not `_start`.

I dropped `argv` which was unused, and renamed `main` to `do_stuff`.

Works?

aleclearmind: I still see `main`

pfez: Now should be ok.

PR #24, opened 2024-03-22
Introduce `switch` promotion test case

andrealinux1:
Add a new source file that introduces the test for the dual case `switch` promotion to `if` statements

### DIFF
--- a/share/revng/test/configuration/revng-qa/for-decompilation.yml
+++ b/share/revng/test/configuration/revng-qa/for-decompilation.yml
@@ -23,6 +23,7 @@ sources:
     members:
       - segments-and-sections.c
       - pretty-ints.c
+      - dual-case-switch.c
   - tags: [simple-executable, native, dynamic, for-decompilation, clang, linked-lists-flags]
     prefix: share/revng/test/tests/analysis/Decompilation/clang/
     members:
--- a/share/revng/test/tests/analysis/Decompilation/x86-64/dual-case-switch.c
+++ b/share/revng/test/tests/analysis/Decompilation/x86-64/dual-case-switch.c
@@ -1,0 +1,37 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+
+#define WEAK __attribute__((weak))
+
+WEAK void report(int code) {
+}
+
+WEAK int f(int arg1) {
+  int return_value;
+  while (arg1 > 0) {
+    if (arg1 == 1) {
+      return_value = 10;
+      report(1);
+      goto LABEL_EXIT_1;
+    } else if (arg1 == 2) {
+      return_value = 20;
+      report(2);
+      goto LABEL_EXIT_2;
+    }
+  }
+
+LABEL_EXIT_1:
+  report(1);
+  goto LABEL_EXIT_COMMON;
+
+LABEL_EXIT_2:
+  report(2);
+
+LABEL_EXIT_COMMON:
+  return return_value;
+}
+
+WEAK void _start() {
+  f(1);
+}


share/revng/test/configuration/revng-qa/for-decompilation.yml
```diff
@@ -27,3 +27,4 @@ sources:
     prefix: share/revng/test/tests/analysis/Decompilation/clang/
     members:
       - linked-lists.c
+      - dual-case-switch.c
```
pfez: Please put this under `x86-64` (like e.g. `segments-and-sections.c`), not  `clang`, unless this is strictly necessary. And if for some reason you really really need `clang` I'd like you to explain why.

share/revng/test/tests/analysis/Decompilation/clang/dual-case-switch.c
```diff
@@ -0,0 +1,35 @@
...
+
+LABEL_EXIT_2:
+  printf("Exit 2");
+
+LABEL_EXIT_COMMON:
+  return return_value;
+}
+
+int main() {
+  int a = 0;
```
pfez: Is this needed?

share/revng/test/tests/analysis/Decompilation/clang/dual-case-switch.c
```diff
@@ -0,0 +1,35 @@
...
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+#include <stdio.h>
+
+int f(int arg1) {
+  int return_value;
+  while (arg1 > 0) {
+    if (arg1 == 1) {
+      return_value = 10;
+      printf("Argument equal to 1");
```
pfez: This should be linked with `nostlib`, so no `printf` please.

I suppose this is just a way to prevent the optimizer to optimize away this call.

Just look into the other files in `x86-64` `for-decompilation` to see how we deal with this.

TLDR: use `__attribute__((weak))` all over all the functions whose call sites you want to prevent the optimizer to mess with or remove.

pfez (changes requested):

share/revng/test/configuration/revng-qa/for-decompilation.yml
andrealinux1: No specific reason, just thought we wanted it done with `clang` since I usually do that when crafting tests. Put it under the `x86-64` folder in https://github.com/revng/revng-qa/commit/7d9d32d70292d147d1eec204ac75342c74c3fd68

share/revng/test/tests/analysis/Decompilation/clang/dual-case-switch.c
andrealinux1: Nope, it was a leftover from a previous implementation. Removed this in https://github.com/revng/revng-qa/commit/7d9d32d70292d147d1eec204ac75342c74c3fd68

share/revng/test/tests/analysis/Decompilation/clang/dual-case-switch.c
andrealinux1: Yeah right, the `printf` was there for the reason you highlighted. Changed it following your suggestion in https://github.com/revng/revng-qa/commit/7d9d32d70292d147d1eec204ac75342c74c3fd68

share/revng/test/tests/analysis/Decompilation/x86-64/dual-case-switch.c
```diff
@@ -0,0 +1,37 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+
+#define WEAK __attribute__((weak))
+
+WEAK void report(int code) {
+}
+
+int f(int arg1) {
```
pfez: Now I wonder why `f` is the only one not marked `WEAK`. Any specific reason?

andrealinux1: No specific reason except my stupidity:) Fixed this in 91cd7fca9e648fe8c5b9286dc3909ed9cf9b83b7

PR #26, opened 2024-05-10
Improve arm64 ABI handling across the board

Cvelth:


revng-test-configure
```diff
@@ -179,7 +179,7 @@ class Variables:
             elif type(self.variables[name]) is not list:
                 error(f"Key {name} is defined with different types")
 
-            self.variables[name] += value
+            self.variables[name] = value + self.variables[name]
```
aleclearmind, line 182: Why?

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
```diff
@@ -31,12 +31,20 @@ mkdir -p "${OUTPUT_DIRECTORY}"
   -o "${OUTPUT_DIRECTORY}/functions"
 
 # Build the binary for runner to `mmap`
+"${TRIPLE}gcc" -c \
+  ${CFLAGS} -O3 -static \
+  -fno-zero-initialized-in-bss \
+  -ffreestanding \
```
aleclearmind: Why?

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
```diff
@@ -31,12 +31,20 @@ mkdir -p "${OUTPUT_DIRECTORY}"
   -o "${OUTPUT_DIRECTORY}/functions"
 
 # Build the binary for runner to `mmap`
+"${TRIPLE}gcc" -c \
+  ${CFLAGS} -O3 -static \
+  -fno-zero-initialized-in-bss \
```
aleclearmind: What was this for?

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
```diff
@@ -31,12 +31,20 @@ mkdir -p "${OUTPUT_DIRECTORY}"
   -o "${OUTPUT_DIRECTORY}/functions"
 
 # Build the binary for runner to `mmap`
+"${TRIPLE}gcc" -c \
+  ${CFLAGS} -O3 -static \
```
aleclearmind: Not sure `-static` makes sense if you're not linking.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
```diff
@@ -31,12 +31,20 @@ mkdir -p "${OUTPUT_DIRECTORY}"
...
+  "${INPUT_DIRECTORY}/setup.c" \
+  -o "${OUTPUT_DIRECTORY}/foreign-executable.o"
 "${TRIPLE}gcc" \
   ${CFLAGS} -O3 -static \
   -Wl,--section-start=.text=0x2000000 \
   -Wl,--section-start=.data=0x3000000 \
-  -fno-zero-initialized-in-bss \
-  "${INPUT_DIRECTORY}/setup.c" \
+  -Wl,--entry=main \
+  -nostdlib -nodefaultlibs \
```
aleclearmind: I guess this significantly reduces the size of the binaries. Correct?

share/revng/test/tests/abi/build_abi_test_binaries_with_msvc.sh
```diff
@@ -48,11 +48,15 @@ timeout 30 ${MSVC_TRIPLE}cl \
...
-timeout 30 ${MSVC_TRIPLE}cl \
+timeout 30 ${MSVC_TRIPLE}cl -c \
   ${MSVC_CFLAGS} -O2 -std:c11 -Zi -GS- \
   "z:\\${INPUT_DIRECTORY}/setup.c" \
-  -Fe"z:\\${OUTPUT_DIRECTORY}/foreign-executable.exe" \
-  -link ${LDFLAGS} /opt:ref,noicf /ignore:4281 /filealign:4096 \
+  -Fo"z:\\${OUTPUT_DIRECTORY}/foreign-executable.obj"
+timeout 30 ${MSVC_TRIPLE}link \
+  ${LDFLAGS} /opt:ref,noicf /filealign:4096 \
+  /ignore:4281 /nodefaultlib /entry:main /subsystem:console \
```
aleclearmind: I guess this significantly reduces the size of the binaries. Correct?

share/revng/test/tests/abi/musl_memory_functions.c
aleclearmind, line 1: I'm OK with importing external implementations, but this is a lot of code, optimized for performance, for functions that could be three lines.
Also `memcpy` can be a call to `memove`.

Also, we might have to mention Rich Felker in the license in some file being installed.

share/revng/test/configuration/revng-qa/architectures.yml
```diff
@@ -46,15 +46,20 @@ tags:
         - -Wl,-Ttext-segment=0x20000
       COMMON_CFLAGS:
         - -DTARGET_arm
+      CLANG_CFLAGS:
+        - --target=arm
```
aleclearmind: If you want to do this, use `"$TRIPLE"` as `--target`. There might be subtle differences, let's use the same triple.

Alternatively, we can have `clang-release` install `armv7a-hardfloat-linux-uclibceabi-clang` symlinked to `clang`, which should be equivalent to `clang --target armv7a-hardfloat-linux-uclibceabi`. This is the appropriate solution, but I'm OK with `--triple` if you leave a `TODO`.

share/revng/test/configuration/revng-qa/architectures.yml
```diff
@@ -87,11 +98,28 @@ tags:
...
       POINTER_SIZE: "64"
       QEMU_NAME: aarch64
+      APPLE_NAME: arm64
       COMMON_CFLAGS:
         - -DTARGET_aarch64
+      CLANG_CFLAGS:
+        - --target=aarch64
+
+  - name: apple-x86-64
+    implies: [x86-64, clang-only]
```
aleclearmind: 1. Does `x86-64` imply `linux`? `apple-x86-64` should not imply `linux`. I'd like an `apple` tag (or `darwin`).

share/revng/test/configuration/revng-qa/architectures.yml
```diff
@@ -87,11 +98,28 @@ tags:
...
+    variables:
+      CLANG_CFLAGS:
+        - --target=x86_64-apple-macos-macho
```
aleclearmind: Where did you get the triple from? I'd expect `darwin` somewhere.

share/revng/test/configuration/revng-qa/compile.yml
```diff
@@ -1,27 +1,29 @@
...
+  - name: clang-only
   - name: simple-executable
   - name: stripped
   - name: with-debug-info
 commands:
   - type: revng-qa.compiled-with-debug-info
     from:
       - type: source
-        filter: linux and simple-executable and !clang
+        filter: linux and simple-executable and !clang-only
```
aleclearmind: If `apple-aarch64` does not have the `linux` tag, this won't be necessary.

share/revng/test/configuration/revng-qa/compile.yml
```diff
@@ -1,27 +1,29 @@
...
     command: ${TRIPLE}gcc $INPUT -o $OUTPUT $GCC_CFLAGS $COMMON_CFLAGS
   - type: revng-qa.compiled-with-debug-info
     from:
       - type: source
-        filter: linux and simple-executable and clang
-    command: ${TRIPLE}clang $INPUT -o $OUTPUT $GCC_CFLAGS $COMMON_CFLAGS
+        filter: linux and simple-executable and clang-only
+    command: clang $INPUT -o $OUTPUT $GCC_CFLAGS $COMMON_CFLAGS $CLANG_CFLAGS
```
aleclearmind: `${TRIPLE}clang` would be the right thing to do, I think.

share/revng/test/configuration/revng-qa/compile.yml
```diff
@@ -1,27 +1,29 @@
...
   - type: revng-qa.compiled-stripped
     from:
       - type: revng-qa.compiled-with-debug-info
+        filter: "!clang-only"
```
aleclearmind: 1. This tag is being abused. If `strip` is ELF specific, let's have an `elf` tag and have `!elf`. Probably `!linux` is the right thing to have here.
2. Does `llvm-strip` support macho?

share/revng/test/tests/abi/build_abi_test_binaries_with_clang.sh
aleclearmind, line 1: How much of this code is duplicated?
Can we merge this script with something else?

revng-test-configure
Cvelth: This way the inherited tag's variables get added _after_ those of the tag it inherits from. And clang cares about order of stuff like `--target=X` if there's more than one specified.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: Because when compiling for arms, the compiler (or is it linker?) is very eager to use `bss` where it can. And the current test toolchain does not support it.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: Yeah, it's a leftover.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: For consistency with the other versions where we try to strip binaries. As far as gcc documentation is concerned, adding this flags means that the compiler is not going to use anything but a couple of basic functions like `memset`, `memcpy` and friends.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: I didn't compare explicitly, but I'd hope so.

share/revng/test/tests/abi/build_abi_test_binaries_with_msvc.sh
Cvelth: Do you want me to build the old ones and check?

share/revng/test/tests/abi/musl_memory_functions.c
Cvelth: After the recent fiasco, I would rather I didn't have to write (and especially debug) those three lines 😅.

That said, this was the first decent-looking implementation I ran into. I don't mind grabbing one from other place if you have something in mind.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: After the first big msvc PR went in, these no longer imply linux. Only `cross-compiler`.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: By fiddling with it until it worked. As far I understand, the only parts we care about are `x86_64` (specifically with underscore), `apple` and `macho`. Everything else doesn't seem to have meaningful impact onto anything I have seen. I did keep `macos` just in case though.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: This is not related to apple though. I just renamed (and reused) the existing `clang` tag to `clang-only`, because it was confusing me.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: I agree, but I that seems non-transparent to me, and it's a lot easier to reproduce when you can just do `any-clang --target=X`.

That being said, now that this finally works, I don't mind any tweaks as long as it keeps working :D

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: 1. Sounds good to me!

2. A quick google search shows that `llvm-objcopy --strip-all` is the way to go.

share/revng/test/tests/abi/build_abi_test_binaries_with_clang.sh
Cvelth: Probably, but I had trouble getting it to work the first place. Now that it does, I don't mind trying to factor out common stuff between these three scripts.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
aleclearmind: Add a comment

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
aleclearmind: Drop it?

share/revng/test/tests/abi/build_abi_test_binaries_with_msvc.sh
aleclearmind: If you can just give a quick test with and without on a single binary, I'm interested.

share/revng/test/tests/abi/musl_memory_functions.c
aleclearmind: 1. Run `clang-format`
2. Drop all optimizations (`#ifdef __GNUC__`), use the "pure C" versions.
3. The "pure C" implementation of `memcpy` is actually `memmove`, have one call the other.

These changes should shrink the file to 40 SLOC.

share/revng/test/configuration/revng-qa/compile.yml
aleclearmind: I'd prefer `${TRIPLE}clang`. It's the convention for cross-compiling.

share/revng/test/configuration/revng-qa/architectures.yml
aleclearmind: Let's refrain from making up new triple, they make me cry.
Use one of these:
```
.orchestra/config/components/toolchain/macos/clang.yml:x86-64: x86_64-apple-darwin11
.orchestra/config/components/toolchain/macos/clang.yml:i686: i686-apple-darwin11
.orchestra/config/components/toolchain/macos/ld64.yml:x86-64: x86_64-apple-darwin11
.orchestra/config/components/toolchain/macos/ld64.yml:i686: i686-apple-darwin11
.orchestra/config/components/toolchain/macos/ld64.yml:arm: arm-apple-darwin11
.orchestra/config/components/toolchain/macos/ld64.yml:aarch64: aarch64-apple-darwin11
```

share/revng/test/tests/abi/build_abi_test_binaries_with_clang.sh
aleclearmind: Go ahead

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: Done.

share/revng/test/tests/abi/build_abi_test_binaries_with_gcc.sh
Cvelth: Done.

share/revng/test/tests/abi/build_abi_test_binaries_with_msvc.sh
Cvelth: `~170KB -> ~72KB` 🎉

share/revng/test/tests/abi/musl_memory_functions.c
Cvelth: Done.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: Left a todo.

share/revng/test/configuration/revng-qa/architectures.yml
Cvelth: These seem to work :tada:

share/revng/test/tests/abi/build_abi_test_binaries_with_clang.sh
Cvelth: Please take a look at the restructured scripts when you have some time.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: Restored.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: 1. Ok, this is _a lot_ more awkward than I expected. Thing is, all the `clang-only` architectures rely on others, so they end up inheriting `linux` (and `elf` too, if we were to add it). And we intentionally disallowed stripping tags. As such, the only viable way to proceed is to add a new tag :sadparrot: Or completely separate architectures. But that'd lead to duplication which IMO is even more dangerous.

2. Done.

aleclearmind: 2. So can we drop using `${TRIPLE}strip` for `!clang-only` and `llvm-objcopy --strip-all` for `clang-only`? I still see both.

Let's review the usages of `clang-only`:
1. When we want to use `clang` (e.g., `share/revng/test/tests/analysis/Decompilation/clang/`). This can be the old `clang` tag.
2. For apple architectures (e.g., `apple-i386`, `apple-x86-64` and `apple-aarch64`) so we use cang. Those tags could simply imply `clang`.
3. In `revng-qa.runtime-abi-tests`/`revng-qa.runtime-abi-tests` to avoid running QEMU on those. We should use `!apple` there.

`clang-only` has a very ad-hoc, hacky and "excluding" semantic, I really want to avoid it.

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
```diff
@@ -209,15 +259,59 @@ commands:
   - type: revng-qa.runtime-abi-tests
     from:
       - type: source
-        filter: for-runtime-abi-tests and !windows
+        filter: for-runtime-abi-tests and !windows and !clang-only
     suffix: /
     command: |-
-      "${INPUT}/build_abi_test_binaries_with_gcc.sh"
+      source "${INPUT}/prepare-functions-binary.sh" &&
```
aleclearmind: No need for `&&`, we should be `set -e`.

share/revng/test/tests/abi/prepare-functions-binary.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
aleclearmind: 1. Always use `#!/usr/bin/env bash`
2. Drop the shebang and `set -euo pipefail`, those belong only to the actual executable script, not on things that are supposed to be `source`'d.

share/revng/test/configuration/revng-qa/compile.yml
Cvelth: 2. Wait, I think I lost you there. Do you mean we should get rid of `${TRIPLE}strip` and just use `llvm-objcopy --strip-all` for everything?

3. Wait, I'm confused about this too. Is it just the _name_ that you don't like? As in `clang-only`->`clang`?

3.3. What do you mean? We still want to run QEMU on everything it supports (which includes aarch64 apple binary).

share/revng/test/tests/abi/prepare-functions-binary.sh
Cvelth: 1. I just thoughtlessly copied it from the old script 😅.
2. Done.

3. On that note, is there an easy way to prevent running it? Something akin to python's
```python
if __name__ == "__main__":
  error_out()
```
?

share/revng/test/configuration/revng-qa/for-runtime-abi-tests.yml
Cvelth: Done.

share/revng/test/tests/abi/prepare-functions-binary.sh
aleclearmind: 3. Not that I know of.

Cvelth: Gotcha, sad. I guess this will have to do then.

share/revng/test/configuration/revng-qa/compile.yml
aleclearmind: 2. > Wait, I think I lost you there. Do you mean we should get rid of ${TRIPLE}strip and just use llvm-objcopy --strip-all for everything?

   Yes, any reason not to?
3. `clang` already existed and had a role which I think we want to preserve.

   > What do you mean? We still want to run QEMU on everything it supports (which includes aarch64 apple binary).
   
   My bad, I got confused.
   The problem is that in the first `revng-qa.runtime-abi-tests`  we use `!clang-only` and we use `prepare-elf-functions`. In the second instance we use `clang-only` + `prepare-macho-functions`. In the third we use `windows` and `prepare-pe-functions`.
   This should be reorganized so we have `linux` (or `elf`) for the first one, `apple` (or `macho`) for the second one and `windows` (or `pe-coff`) for the third one.
   Right now we have `!windows and !clang-only`, `!windows and clang-only` and then `windows`, very confusing.

I understand I'm no expressing myself too clearly, if you want I can give a shot at it.

Cvelth: Done, please take a look!

aleclearmind:
Merged in 62e18aa. Thanks.

PR #34, opened 2026-05-11
Floating Point Frontend

andrealinux1:
Add a binary in revng-qa used to check for the emergence of the softfloat calls in the lifted IR.

### DIFF
--- a/share/revng/test/configuration/revng-qa/for-floating-point.yml
+++ b/share/revng/test/configuration/revng-qa/for-floating-point.yml
@@ -1,0 +1,29 @@
+#
+# This file is distributed under the MIT License. See LICENSE.md for details.
+#
+
+tags:
+  - name: for-floating-point
+
+  # We force VFPv3 with hardfloat calling convention on armv7a, in order to
+  # avoid the toolchain emitting calls to the libgcc helpers for double
+  # precision FP.
+  - name: arm-double-vfp
+    variables:
+      GCC_CFLAGS:
+        - -mfpu=vfpv3
+        - -mfloat-abi=hard
+
+sources:
+  - tags: [simple-executable, static, nostdlib, nostartfiles, for-floating-point]
+    repeat-for:
+      - [arm, arm-double-vfp]
+      - [mips]
+      - [mipsel]
+      - [x86-64]
+      - [s390x]
+      - [i386]
+      - [aarch64]
+    prefix: share/revng/test/tests/floating-point
+    members:
+      - floating-point.c
--- a/share/revng/test/tests/floating-point/floating-point.c
+++ b/share/revng/test/tests/floating-point/floating-point.c
@@ -1,0 +1,120 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ */
+
+#include <stdint.h>
+
+typedef float float32_t;
+typedef double float64_t;
+
+_Static_assert(sizeof(float32_t) == 4, "float32_t must be 4 bytes wide");
+_Static_assert(sizeof(float64_t) == 8, "float64_t must be 8 bytes wide");
+
+#define WEAK __attribute__((weak))
+
+/*
+ * The operands and results are globals: the compiler cannot constant-fold the
+ * floating-point work away, so the softfloat calls reliably emerge in the
+ * lifted IR. The matching `CHECK` directives live in the `.filecheck` file in
+ * the `revng` repository.
+ */
+float64_t float64_a, float64_b, float64_result;
+float32_t float32_a, float32_b, float32_result;
+int32_t int32_a, int32_result;
+int64_t int64_a, int64_result;
+
+WEAK void add_float64(void) {
+  float64_result = float64_a + float64_b;
+}
+
+WEAK void add_float32(void) {
+  float32_result = float32_a + float32_b;
+}
+
+WEAK void mul_float64(void) {
+  float64_result = float64_a * float64_b;
+}
+
+WEAK void mul_float32(void) {
+  float32_result = float32_a * float32_b;
+}
+
+WEAK void div_float64(void) {
+  float64_result = float64_a / float64_b;
+}
+
+WEAK void div_float32(void) {
+  float32_result = float32_a / float32_b;
+}
+
+WEAK void compare_float64(void) {
+  int32_result = float64_a > float64_b;
+}
+
+WEAK void compare_float32(void) {
+  int32_result = float32_a > float32_b;
+}
+
+WEAK void convert_int32_t_to_float64(void) {
+  float64_result = int32_a;
+}
+
+WEAK void convert_float64_to_int32_t(void) {
+  int32_result = float64_a;
+}
+
+WEAK void convert_int32_t_to_float32(void) {
+  float32_result = int32_a;
+}
+
+WEAK void convert_float32_to_int32_t(void) {
+  int32_result = float32_a;
+}
+
+/*
+ * On 32-bit architectures, the toolchain lowers 64-bit-integer <-> floating
+ * point conversions through libgcc helpers, which then use the underlying
+ * 32-bit primitives, which are already tested.
+ * Their `CHECK-64` directives in the `.filecheck` file are gated on the same
+ * architectures.
+ */
+#if defined(__SIZEOF_POINTER__) && __SIZEOF_POINTER__ == 8
+
+WEAK void convert_int64_t_to_float64(void) {
+  float64_result = int64_a;
+}
+
+WEAK void convert_float64_to_int64_t(void) {
+  int64_result = float64_a;
+}
+
+WEAK void convert_int64_t_to_float32(void) {
+  float32_result = int64_a;
+}
+
+WEAK void convert_float32_to_int64_t(void) {
+  int64_result = float32_a;
+}
+
+#endif
+
+void _start(void) {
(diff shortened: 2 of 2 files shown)


share/revng/test/configuration/revng-qa/for-floating-point.yml
```diff
@@ -0,0 +1,36 @@
...
+      GCC_CFLAGS:
+        - -mfpu=vfpv3
+        - -mfloat-abi=hard
+
+  # We force `mips2` in order to use the `trunc.w.d` / `trunc.w.s` single opcode
+  # to round toward zero, and discard the use of `FCSR`.
+  - name: mips-mips2
+    variables:
+      GCC_CFLAGS:
+        - -mips2
```
aleclearmind: 1. I've no idea what FCSR is and why it's a problem.
2. > use the `trunc.w.d` / `trunc.w.s` single opcode

   What's the alternative? What happens if you don't do this and why is that a problem?

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ *
+ *
```
aleclearmind: Split here the commit

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
...
+ * compiler constant-fold the FP work back to integer arithmetic
+ * at -O2, we add a `+ 0.5` (or `+ 0.5f`).
+ *
+ */
+
+__attribute__((weak)) int add64(int a, int b) {
+  // CHECK-LABEL: @add64(
+  // CHECK:       @float64_add(
+  return (int)((double)a + (double)b + 0.5);
+  // CHECK-NOT: call{{.*}}@float[0-9]+_{{(mul|div|sub|sqrt|rem)}}(
```
aleclearmind: Drop, I don't see the point.

I'd add

```
// CHECK: }
```

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
+/*
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ *
+ *
+ * Where the simple form `(int)((double)a OP (double)b)` would let the
+ * compiler constant-fold the FP work back to integer arithmetic
+ * at -O2, we add a `+ 0.5` (or `+ 0.5f`).
```
aleclearmind: Mmmh, let's simplify things: let's use globals. No casts.

```c
#define WEAK __attribute__((weak))

double double1;
double double2;
double double3;

WEAK void add64(void) {
  double3 = double1 + double1;
}
```

Also, let's have and always use these types:

```c
typedef float float32_t;
typedef double float64_t;

_Static_assert(sizeof(float32_t) == 4);
_Static_assert(sizeof(float64_t) == 8);
```

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+__attribute__((weak)) long long conv_f64_to_i64(long long a) {
+  // CHECK-64-LABEL: @conv_f64_to_i64(
+  // CHECK-64:       @float64_to_int64{{(_scalbn|_round_to_zero)?}}(
+  return (long long)((double)a + 0.5);
+  // CHECK-64-NOT: call{{.*}}@float[0-9]+_{{(mul|div|sub|sqrt|rem)}}(
+}
+
+__attribute__((weak)) long long conv_i64_to_f32(long long a) {
```
aleclearmind: Never abbreviate, expand all abbreviations [B.1](https://pad.rev.ng/s/SJNUet_ku#B-Language-agnostic-practices)

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
...
+ * This file is distributed under the MIT License. See LICENSE.md for details.
+ *
+ *
+ * Where the simple form `(int)((double)a OP (double)b)` would let the
+ * compiler constant-fold the FP work back to integer arithmetic
+ * at -O2, we add a `+ 0.5` (or `+ 0.5f`).
+ *
+ */
+
+__attribute__((weak)) int add64(int a, int b) {
```
aleclearmind: `add_float64`

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+__attribute__((weak)) int conv_i32_to_f64(int a) {
+  // CHECK-LABEL: @conv_i32_to_f64(
+  // CHECK:       @{{int(32|64)}}_to_float64{{(_scalbn)?}}(
+  return (int)((double)a + 0.5);
+  // CHECK-NOT: call{{.*}}@float[0-9]+_{{(mul|div|sub|sqrt|rem)}}(
+}
+
+__attribute__((weak)) int conv_f64_to_i32(int a) {
```
aleclearmind: `convert_float64_to_int32_t`

Always use size-explicit types: `int` -> `int32_t`.

share/revng/test/tests/floating-point/floating-point.c
```diff
@@ -0,0 +1,152 @@
...
+  mul32(1, 2);
+  div64(1, 2);
+  div32(1, 2);
+  cmp64(1, 2);
+  cmp32(1, 2);
+  conv_i32_to_f64(1);
+  conv_f64_to_i32(1);
+  conv_i32_to_f32(1);
+  conv_f32_to_i32(1);
+#if defined(__SIZEOF_POINTER__) && __SIZEOF_POINTER__ == 8
```
aleclearmind, line 114: Why?

share/revng/test/tests/floating-point/floating-point.c
aleclearmind, line 1: We need to split this file. We can't have LLVM IR `CHECK` directives in `revng-qa`.

Create a file in the same place in `revng` with the `.filecheck` suffix containing only the `CHECK` directives.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: Split here the commit? You mean having one commit for the `yml` driver and one for the source? Or you meant to just split the comment? I split the *comment* for the moment being

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: Ok, I took your various suggestions in `[Add floating-point test]`.

- I took the liberty of using `float32_a`, `int32_result` as variable names, in order to be less confusing, let me know if you prefer a different convention.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: This file should be now clear of abbreviations.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: Changed this and all the related names.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: Expanded the name, since we now use the globals the types are gone from the body and from the signatures, it's a bit strange, but the variable names recall the types in some way.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: I added a comment explaining this on top of the first `#if defined`. I'll report it there:
 * On 32-bit architectures, the toolchain lowers 64-bit-integer <-> floating point conversions through libgcc helpers, which then use the underlying 32-bit primitives, which are already tested.
 * Their `CHECK-64` directives in the `.filecheck` file are gated on the same architectures.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: I removed the `CHECK` directives from this file, and moved them in the dedicated `.filecheck` file as you suggested.

share/revng/test/tests/floating-point/floating-point.c
andrealinux1: I dropped the `CHECK-NOT` directives and added the `CHECK: }` for each block, in the `.filecheck` file in the `revng` repo.

share/revng/test/tests/floating-point/floating-point.c
aleclearmind: Comment, sorry

share/revng/test/tests/floating-point/floating-point.c
aleclearmind: If you're confident you fixed the problem and need no further comment/check from me, please mark as solved

share/revng/test/configuration/revng-qa/for-floating-point.yml
andrealinux1: Due to the restructuring of the tests, this is not an immediate problem anymore for the tests. As mentioned, I saved all the details to reproduce this.

I removed the `mips2` compilation flag in `[Add floating-point test]`.

aleclearmind:
Merged in e650b9f. Thanks.

